PR #30, opened 2022-06-02
Use deal function to set ETH/ERC20 balance

NIC619:
Turns out there's already `deal` function to set ERC20 balance in `forge-std` lib


pilagod (approved):
This is convenient. Not sure how it achieves this based on the fact that each token has arbitrary storage layout for balance 🧐

NIC619:
It records the storage slot accessed when querying `balanceOf(address)` or `totalSupply()`. Very clever.🧐

PR #32, opened 2022-06-06
[DO NOT MERGE] Try add stETH pool

NIC619:


NIC619:
I missed that due to how `balanceOf` is calculated in a rebase token, it's possible that it fails Spender's balance check after transfer.

PR #33, opened 2022-06-06
Add Spender tests for non standard ERC20 tokens

NIC619:
- Add Spender tests for Deflationary token, NoReturn token and NoRevert token

### DIFF
--- a/contracts/test/Spender.t.sol
+++ b/contracts/test/Spender.t.sol
@@ -220,13 +220,13 @@ contract SpenderTest is BalanceUtil {
         spender.spendFromUser(user, address(lon), 100);
     }
 
-    function testCannotSpendFromUserWithNoReturnValueToken() public {
+    function testCannotSpendFromUserInsufficientBalance_NoReturnValueToken() public {
         uint256 userBalance = noReturnERC20.balanceOf(user);
         vm.expectRevert("Spender: ERC20 transferFrom failed");
         spender.spendFromUser(user, address(noReturnERC20), userBalance + 1);
     }
 
-    function testCannotSpendFromUserWithReturnFalseToken() public {
+    function testCannotSpendFromUserInsufficientBalance_ReturnFalseToken() public {
         uint256 userBalance = noRevertERC20.balanceOf(user);
         vm.expectRevert("Spender: ERC20 transferFrom failed");
         spender.spendFromUser(user, address(noRevertERC20), userBalance + 1);
@@ -243,6 +243,14 @@ contract SpenderTest is BalanceUtil {
         spender.spendFromUser(user, address(lon), 100);
 
         assertEq(lon.balanceOf(address(this)), 100);
+    }
+
+    function testSpendFromUserWithNoReturnValueToken() public {
+        assertEq(noReturnERC20.balanceOf(address(this)), 0);
+
+        spender.spendFromUser(user, address(noReturnERC20), 100);
+
+        assertEq(noReturnERC20.balanceOf(address(this)), 100);
     }
 
     /*********************************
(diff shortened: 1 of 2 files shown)


contracts/test/Spender.t.sol
```diff
@@ -202,6 +220,23 @@ contract SpenderTest is Test {
         spender.spendFromUser(user, address(lon), 100);
     }
 
+    function testCannotSpendFromUserWithNoReturnValueToken() public {
+        uint256 userBalance = noReturnERC20.balanceOf(user);
+        vm.expectRevert("Spender: ERC20 transferFrom failed");
+        spender.spendFromUser(user, address(noReturnERC20), userBalance + 1);
```
charlesjhongc, line 226: Why setting transfer amount to `userBalance + 1`?

NIC619: To cause the transaction to revert so Spender can catch it:
`require(callSucceed, "Spender: ERC20 transferFrom failed");`

NIC619: Maybe we can one more test case which succeeds in transferring a `noReturnERC20` token.

charlesjhongc: Ok, I think it's about the naming. `testCannotSpendFromUserWithNoReturnValueToken` looks like spender cannot spend token without return value to me. Perhaps these can be merged as one test case `testCannotSpendFromUserInsufficientBalance` and test special tokens in `testSpendFromUser` as well.

contracts/test/Spender.t.sol
```diff
@@ -51,8 +63,14 @@ contract SpenderTest is Test {
         // Mint 10k tokens to user
         lon.mint(user, 10000 * 1e18);
         // User approve AllowanceTarget
-        vm.prank(user);
+        vm.startPrank(user);
         lon.approve(address(allowanceTarget), type(uint256).max);
+        // Set user's mock tokens balance and approve
+        for (uint256 j = 0; j < tokens.length; j++) {
+            setERC20Balance(address(tokens[j]), user, 100);
```
pilagod, line 70: Could we use `deal` here?

And if this is possible, maybe `StrategySharedSetup.sol:122` can do the same refactoring

contracts/test/Spender.t.sol
NIC619: Good point. Will rename and add the tests. Though I will prefer separating them into two cases: `testCannotSpendFromUserInsufficientBalance_NoReturnValueToken`, `testCannotSpendFromUserInsufficientBalance_NoRevertToken` and `testSpendFromUserWithNoReturnValueToken`.

contracts/test/Spender.t.sol
NIC619: `deal` would work but act like `setERC20BalanceRaw` so that's why I keep `setERC20Balance` as it can handle the `decimal` for us.

pilagod (approved):

contracts/test/Spender.t.sol
pilagod: Not sure but I think decimal is more for display purpose. Decimal is even not required in ERC20 spec. Maybe our test doesn't need to care too much about decimal.

NIC619: For RFQ/LimitOrder, it should work fine without decimal. But for AMM it probably matters unless we only test with small amount?

charlesjhongc: I support this wrapping for decimal since conventional amount is not using `uint256` as unit but instead using weighted unit. This matters when it comes to price as @NIC619 said. For example `1 DAI approx equals to 1 USDT value` is actually `1 DAI euqals to 1e12 USDT` since DAI has decimal in 18 while USDT is in 6. I think either way we need a util to translate it.

charlesjhongc (approved):

pilagod (approved):

PR #34, opened 2022-06-09
[DO NOT MERGE] Demonstration of removing Spender balance check

NIC619:
- Remove balance check in Spender so rebase tokens and fee charging tokens can be supported
- Update AMM to use transferred amount to continue swapping
- Update RFQ to check transferred amount and revert if amount mismatched


charlesjhongc:
Close PR but keep the branch.

PR #35, opened 2022-06-09
remove subsidy from AMMWrapper

charlesjhongc:
Remove AMMWrapping subsidy design and refactor fee logic. Allow feeFactor to be set by relayer in relayed case. Overwrite feeFactor using default one in general user case.

### DIFF
--- a/contracts/AMMWrapper.sol
+++ b/contracts/AMMWrapper.sol
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: MIT
 pragma solidity 0.7.6;
 pragma abicoder v2;
 
--- a/contracts/AMMWrapper.sol
+++ b/contracts/AMMWrapper.sol
@@ -1,4 +1,5 @@
 pragma solidity 0.7.6;
+pragma abicoder v2;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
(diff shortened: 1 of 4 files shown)


contracts/AMMWrapper.sol
```diff
@@ -196,82 +175,36 @@ contract AMMWrapper is IAMMWrapper, ReentrancyGuard, BaseLibEIP712, SignatureVal
...
+        txMetaData.transactionHash = _verify(_order, _sig);
 
-        _prepare(order, internalTxData);
+        _prepare(_order, internalTxData);
 
-        // minAmount = makerAssetAmount * (10000 - subsidyFactor) / 10000
-        uint256 _minAmount = order.makerAssetAmount.mul((BPS_MAX.sub(txMetaData.subsidyFactor))).div(BPS_MAX);
-        (txMetaData.source, txMetaData.receivedAmount) = _swap(order, internalTxData, _minAmount);
+        // amountOutMin = makerAssetAmount * (10000 + feeFactor) / 10000
+        uint256 _amountOutMin = _order.makerAssetAmount.mul((LibConstant.BPS_MAX.add(feeFactor))).div(LibConstant.BPS_MAX);
```
pilagod, line 201: Would it be harder to settle when we require an additional fee on swap out amount? 

I think maybe ensuring to swap out at least `makerTokenAmount` is enough. As long as maker knows we charge fee, it's reasonable to give maker a slightly less amount.

charlesjhongc: I was worried about user get the wrong idea when they receive less than the quote. But this can be adjusted in the front end as well. Let's check the current front end policy and align with it.

contracts/AMMWrapper.sol
```diff
@@ -336,9 +334,17 @@ contract AMMWrapper is IAMMWrapper, ReentrancyGuard, BaseLibEIP712, SignatureVal
...
+            // shouldCollectFee = ((receivedAmount - makerAssetAmount) / receivedAmount) > (feeFactor / 10000)
+            bool shouldCollectFee = _txMetaData.receivedAmount.sub(_order.makerAssetAmount).mul(LibConstant.BPS_MAX) >
+                feeFactor.mul(_txMetaData.receivedAmount);
+            if (shouldCollectFee) {
+                // settleAmount = receivedAmount * (1 - feeFactor) / 10000
+                settleAmount = _txMetaData.receivedAmount.mul(LibConstant.BPS_MAX.sub(feeFactor)).div(LibConstant.BPS_MAX);
+            } else {
+                settleAmount = _order.makerAssetAmount;
+            }
+        }
```
pilagod, line 360: Use an example to clarify the logic, suppose fee factor is 5% and make amount is 100, then

(1) swap out 99 -> revert
(2) swap out 103 -> maker takes 100, our AMM takes 3
(3) swap out 110 -> maker takes 104.5, our AMM takes 5.5

So when swapped out surplus is less than our fee factor, we would still collect fee, but it would be less than fee factor set on contract, and maker would receive exactly the maker amount. For maker to receive more than maker amount, the swapped out surplus must be more than fee factor.

Is this understanding correct?

NIC619: I think we should adjust the frontend to show the amount after fee is charged instead of doing the calculation and determine if we should charge fee in contract, i.e., we should charge as long as the AMM swap succeeded.

Because if the `makerAssetAmount` is supplied not by us but by user, user can fine tune the `makerAssetAmount` so we can never charge fee from him.

pilagod: Sounds reasonable.

charlesjhongc:
Close this stale one. Handled by #55 now.

PR #36, opened 2022-06-10
Add PermanentStorage tests

NIC619:


charlesjhongc (approved):

contracts/test/PermanentStorage.t.sol
```diff
@@ -0,0 +1,428 @@
...
+            address underlyingCoinA = DEFAULT_CURVE_POOL_UNDERLYING_COINS[i];
+            for (uint256 j = i; j < DEFAULT_CURVE_POOL_UNDERLYING_COINS.length; j++) {
+                address underlyingCoinB = DEFAULT_CURVE_POOL_UNDERLYING_COINS[j];
+                (int128 takerAssetIndex, int128 makerAssetIndex, uint16 swapMethod, bool supportGetDx) = permanentStorage.getCurvePoolInfo(
+                    DEFAULT_CURVE_POOL_ADDRESS,
+                    underlyingCoinA,
+                    underlyingCoinB
+                );
+                assertGt(takerAssetIndex, 0);
+                assertGt(makerAssetIndex, 0);
```
pilagod, line 405: Q: Not sure why the indexes are all 0, I thought they would be increasing?

NIC619: It would more complex to check the exact index so I use `assert greater than` instead to at least check that they are written to and hence not returning a default value `0`.

pilagod: Oh I miss the `Gt` keyword 🤦‍♂️ No problem 💪

pilagod (approved):

PR #37, opened 2022-06-13
Add TreasuryVester tests

NIC619:


charlesjhongc (approved):

contracts/test/TreasuryVester.t.sol
```diff
@@ -0,0 +1,272 @@
...
+        lon.approve(address(treasuryVesterFactory), DEFAULT_VESTING_AMOUNT);
+        treasuryVester = TreasuryVester(
+            treasuryVesterFactory.createVester(
+                recipient,
+                DEFAULT_VESTING_AMOUNT,
+                DEFAULT_VESTING_BEGIN_TIMESTAMP,
+                DEFAULT_VESTING_CLIFF_TIMESTAMP,
+                DEFAULT_VESTING_END_TIMESTAMP
+            )
+        );
```
pilagod, line 178: nit: Maybe it's worth a util function to setup a valid vester

pilagod (dismissed):

contracts/test/TreasuryVester.t.sol
NIC619: Fixed in 66a1b2f

pilagod: @NIC619 I think approve can also include in the util. What do you think about it?

NIC619: I feel like it's not worth the readability improvement as much as creating a default vester? Also the function name would probably need to be changed to `_createVesterWithDefaultValuesAndApproveTreasuryFactory` as we want to be as explicit as possible in our tests and I don't not a fan of the name.😂 (maybe there's better name🤔)

pilagod: Actually what my original thought is not that complicate 😆 I was thinking something like

```solidity
function setupTreasuryVester(recipient, amount, begin, cliff, end) {
    lon.approve(address(treasuryVesterFactory), amount);
    return TreasuryVester(
            treasuryVesterFactory.createVester(
                recipient,
                amount,
                begin,
                cliff,
                end
            )
        );
}
```

But this is just a minor suggestion that we can take it not so serious. The original version it's ok for me too. We can just choose a preferable way for us to accomplish it 👍

pilagod: And I come up another idea that if we separate contract test cases into different files (as we discussed before), each file can have their custom before each hook. 

For these cases based on valid TreasuryVester, we can simply move more setup steps into setup function.

Just idea sharing, not to do right now 😆

NIC619: Yeah breaking tests into different files is definitely an good option if the test file grows too large.

NIC619: Let's settle with current version?

pilagod: OK, no problem 👍

pilagod (approved):

NIC619:
Shoot. I accidentally commit and push LONStaking tests to this branch. Forced push to remove the commits...

pilagod (approved):

charlesjhongc (approved):

PR #38, opened 2022-06-14
Add LONStaking fuzzing tests

NIC619:
- Change default fuzz runs from `50000` to 1000 so the tests won't cost too much time
- `stake`:
    - simulate multiple stakers staking with different amount
- `redeem`:
    - simulate multiple stakers staking with different amount and each redeem with different amount
    - `OneByOne` scenario: one staker stake, unstake and redeem after another staker
    - `AllAtOnce` scenario: all stakers stake and unstake at once and then all redeem at once
- `rageExit`:
    - simulate multiple stakers staking with different amount and rage exit
    - `OneByOne` scenario: one staker stake, unstake and rage exit after another staker
    - `AllAtOnce` scenario: all stakers stake and unstake at once and then all rage exit at once

### DIFF
--- a/contracts/test/LONStaking.t.sol
+++ b/contracts/test/LONStaking.t.sol
@@ -31,6 +31,7 @@ contract LONStakingTest is Test {
     address other = vm.addr(otherPrivateKey);
     address upgradeAdmin = address(0x133701);
     address spender = address(0x133702);
+    address fuzzingUserStartAddress = address(0x133703);
 
     Lon lon = new Lon(address(this), address(this));
     xLON xLon;
@@ -253,6 +254,15 @@ contract LONStakingTest is Test {
         lonStaking.stake(DEFAULT_STAKE_AMOUNT);
     }
 
+    function _stake(address staker, uint256 stakeAmount) internal {
+        vm.startPrank(staker);
+        if (lon.allowance(staker, address(lonStaking)) == 0) {
+            lon.approve(address(lonStaking), type(uint256).max);
+        }
+        lonStaking.stake(stakeAmount);
+        vm.stopPrank();
+    }
+
     function _stakeAndValidate(address staker, uint256 stakeAmount) internal {
         BalanceSnapshot.Snapshot memory stakerLon = BalanceSnapshot.take(staker, address(lon));
         BalanceSnapshot.Snapshot memory lonStakingLon = BalanceSnapshot.take(address(lonStaking), address(lon));
@@ -292,7 +302,7 @@ contract LONStakingTest is Test {
         }
 
         for (uint256 i = 0; i < stakeAmounts.length; i++) {
-            address staker = address(uint256(user) + i);
+            address staker = address(uint256(fuzzingUserStartAddress) + i);
             uint256 stakeAmount = stakeAmounts[i];
             lon.mint(staker, stakeAmount);
             _stakeAndValidate(staker, stakeAmount);
@@ -315,10 +325,11 @@ contract LONStakingTest is Test {
 
         // Make initial big enough deposit so LON amount will not increase dramatically relative to xLON amount due to buyback
         // and hence result in later staker getting zero shares
+        lon.mint(address(0x5566), 10_000_000e18);
         _stake(address(0x5566), 10_000_000e18); // stake 10m LON
 
         for (uint256 i = 0; i < stakeAmounts.length; i++) {
-            address staker = address(uint256(user) + i);
+            address staker = address(uint256(fuzzingUserStartAddress) + i);
             uint256 stakeAmount = stakeAmounts[i];
             lon.mint(staker, stakeAmount);
             _stakeAndValidate(staker, stakeAmount);
@@ -339,10 +350,11 @@ contract LONStakingTest is Test {
 
         // Make initial big enough deposit so LON amount will not increase dramatically relative to xLON amount due to buyback
         // and hence result in later staker getting zero shares
+        lon.mint(address(0x5566), 10_000_000e18);
         _stake(address(0x5566), 10_000_000e18); // stake 10m LON
 
         // First batch of users stake
-        address firstBatchUsersAddressStart = user;
+        address firstBatchUsersAddressStart = fuzzingUserStartAddress;
         for (uint256 i = 0; i < stakeAmounts.length; i++) {
             address firstBatchUser = address(uint256(firstBatchUsersAddressStart) + i);
             uint256 stakeAmount = stakeAmounts[i];
@@ -351,7 +363,7 @@ contract LONStakingTest is Test {
         }
         simulateBuyback(buybackAmount);
         // Second batch of users stake
-        address secondBatchUsersAddressStart = address(uint256(user) + stakeAmounts.length);
+        address secondBatchUsersAddressStart = address(uint256(fuzzingUserStartAddress) + stakeAmounts.length);
         for (uint256 i = 0; i < stakeAmounts.length; i++) {
             address firstBatchUser = address(uint256(firstBatchUsersAddressStart) + i);
             address secondBatchUser = address(uint256(secondBatchUsersAddressStart) + i);
@@ -432,18 +444,6 @@ contract LONStakingTest is Test {
     /*********************************
      *        Test: unstake          *
      *********************************/
-
-    function _stake(address staker, uint256 stakeAmount) internal {
-        if (lon.balanceOf(staker) < stakeAmount) {
-            lon.mint(staker, stakeAmount);
-        }
-        vm.startPrank(staker);
-        if (lon.allowance(staker, address(lonStaking)) == 0) {
-            lon.approve(address(lonStaking), type(uint256).max);
-        }
-        lonStaking.stake(stakeAmount);
-        vm.stopPrank();
-    }
 
     function testCannotUnstakeWithZeroAmount() public {
         vm.expectRevert("no share to unstake");
@@ -574,6 +574,7 @@ contract LONStakingTest is Test {
         vm.assume(stakeAmount <= lon.cap());
         vm.assume(stakeAmount.add(lon.totalSupply()) <= lon.cap());
 
+        lon.mint(user, stakeAmount);
         _stake(user, stakeAmount);
         vm.prank(user);
         lonStaking.unstake();
@@ -614,9 +615,10 @@ contract LONStakingTest is Test {
         }
 
         for (uint256 i = 0; i < stakeAmounts.length; i++) {
-            address staker = address(uint256(user) + i);
+            address staker = address(uint256(fuzzingUserStartAddress) + i);
             uint256 stakeAmount = stakeAmounts[i];
             uint256 redeemAmount = redeemAmounts[i];
+            lon.mint(staker, stakeAmount);
             _stake(staker, stakeAmount);
             vm.prank(staker);
             lonStaking.unstake();
@@ -643,8 +645,9 @@ contract LONStakingTest is Test {
 
         // All stake and unstake
         for (uint256 i = 0; i < stakeAmounts.length; i++) {
-            address staker = address(uint256(user) + i);
+            address staker = address(uint256(fuzzingUserStartAddress) + i);
             uint256 stakeAmount = stakeAmounts[i];
+            lon.mint(staker, stakeAmount);
             _stake(staker, stakeAmount);
             vm.prank(staker);
             lonStaking.unstake();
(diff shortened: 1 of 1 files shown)


contracts/test/LONStaking.t.sol
```diff
@@ -248,17 +249,51 @@ contract LONStakingTest is Test {
...
+        vm.assume(stakeAmount > 0);
+        vm.assume(stakeAmount <= lon.cap());
+        vm.assume(stakeAmount.add(lon.totalSupply()) <= lon.cap());
+
+        lon.mint(user, stakeAmount);
+        _stakeAndValidate(user, stakeAmount);
+    }
+
+    function testFuzz_StakeMultiple(uint80[16] memory stakeAmounts) public {
+        // LON cap lies between 2**87 and 2**88, setting upper bound of stake amount to 2**80 - 1 so stake amount will not exceed cap
```
NIC619, line 282: Setting type of `stakeAmount` to `uint80` so it's max value is `2**80 - 1` so all generated numbers will be less than LON cap. This is to save time from generating invalid `stakeAmount` because now in each run we generate a total of 16 numbers instead of 1, one invalid number would result in generating 16 new numbers.

contracts/test/LONStaking.t.sol
```diff
@@ -433,80 +471,302 @@ contract LONStakingTest is Test {
...
+        _stake(user, stakeAmount);
+        vm.prank(user);
+        lonStaking.unstake();
+
+        vm.warp(block.timestamp + COOLDOWN_SECONDS + 1);
+
+        _redeemAndValidate(user, redeemAmount);
+    }
+
+    function _determineStakeRedeemAmount(uint80[2][16] memory stakeAndRedeemAmounts)
```
NIC619, line 523: Generating two different arrays, `stakeAmounts` and `redeemAomunts`, and making sure `stakeAmounts[i] > redeemAmounts[i]` take too much time so changed to just generating a pair of number and assigning bigger one to `stakeAmount`.

contracts/test/LONStaking.t.sol
```diff
@@ -433,80 +471,302 @@ contract LONStakingTest is Test {
...
+                stakeAmounts[i] = stakeAndRedeemAmounts[i][1];
+                redeemAmounts[i] = stakeAndRedeemAmounts[i][0];
+            }
+        }
+    }
+
+    function testFuzz_RedeemPartialMultiple_OneByOne(uint80[2][16] memory stakeAndRedeemAmounts) public {
+        // LON cap lies between 2**87 and 2**88, setting upper bound of stake amount to 2**80 - 1 so stake amount will not exceed cap
+        uint256 totalLONAmount = lon.totalSupply();
+        (uint256[16] memory stakeAmounts, uint256[16] memory redeemAmounts) = _determineStakeRedeemAmount(stakeAndRedeemAmounts);
```
NIC619, line 542: Technically there are changes where `stakeAmount == redeemAmount` so it's not exactly partial redeem in every cases.

NIC619: We can check `stakeAmount > redeemAmount` but again it will take many time regenerating the whole 16 numbers.

contracts/test/LONStaking.t.sol
```diff
@@ -248,17 +249,51 @@ contract LONStakingTest is Test {
...
+    function testFuzz_Stake(uint256 stakeAmount) public {
+        vm.assume(stakeAmount > 0);
+        vm.assume(stakeAmount <= lon.cap());
+        vm.assume(stakeAmount.add(lon.totalSupply()) <= lon.cap());
+
+        lon.mint(user, stakeAmount);
+        _stakeAndValidate(user, stakeAmount);
+    }
+
+    function testFuzz_StakeMultiple(uint80[16] memory stakeAmounts) public {
```
NIC619, line 281: The number `16` is not justified. It can be increased and depends on how much time are we willing to spend on generating valid inputs.

pilagod: So foundry will base on the argument type to generate valid inputs, including array with multiple dimension, right?

contracts/test/LONStaking.t.sol
```diff
@@ -248,17 +249,51 @@ contract LONStakingTest is Test {
...
+            vm.assume(stakeAmounts[i] > 0);
+            totalLONAmount = totalLONAmount.add(stakeAmounts[i]);
+            vm.assume(totalLONAmount <= lon.cap());
+        }
+
+        for (uint256 i = 0; i < stakeAmounts.length; i++) {
+            address _user = address(uint256(user) + i);
+            uint256 stakeAmount = stakeAmounts[i];
+            lon.mint(_user, stakeAmount);
+            vm.prank(_user);
```
pilagod, line 294: I think this prank is no use, right?

contracts/test/LONStaking.t.sol
```diff
@@ -332,6 +367,9 @@ contract LONStakingTest is Test {
      *********************************/
 
     function _stake(address staker, uint256 stakeAmount) internal {
+        if (lon.balanceOf(staker) < stakeAmount) {
+            lon.mint(staker, stakeAmount);
+        }
```
pilagod, line 372: I feel this is a little implicit, maybe we could derive another function `mintToStake`, `mustStake` or other good namings.

```solidity
function mustStake(staker, stakeAmount) {
    if (lon.balanceOf(staker) < stakeAmount) {
        lon.mint(staker, stakeAmount);
    }
    _stake(staker, stakeAmount)
}
```

contracts/test/LONStaking.t.sol
NIC619: Yes!

contracts/test/LONStaking.t.sol
```diff
@@ -433,80 +471,302 @@ contract LONStakingTest is Test {
...
+            uint256 redeemAmount = redeemAmounts[i];
+            vm.assume(stakeAmount > 0);
+            vm.assume(redeemAmount > 0);
+            vm.assume(stakeAmount > redeemAmount);
+            totalLONAmount = totalLONAmount.add(stakeAmount);
+            vm.assume(totalLONAmount <= lon.cap());
+        }
+
+        for (uint256 i = 0; i < stakeAmounts.length; i++) {
+            address _user = address(uint256(user) + i);
```
pilagod, line 554: nit: Maybe we could call this redeemer? Only this variable with `_` prefix is a little inconsistent 😆

foundry.toml
```diff
@@ -15,7 +15,7 @@ initial_balance = '0xffffffffffffffffffffffff'                # the initial bala
 gas_limit = 9223372036854775807                               # the gas limit in tests
 gas_price = 0                                                 # the gas price (in wei) in tests
 
-fuzz_runs = 50000                                             # the number of fuzz runs for tests
+fuzz_runs = 1000                                              # the number of fuzz runs for tests
```
pilagod, line 18: Q: I'm curious about that this number is for only valid inputs, or total runs including invalid inputs?

pilagod (commented):
I have no experience in fuzz testing before. I feel that in order to do fuzz testing, we need to add logic to abstract test code further (like line 370, 475), which might lost clarity and cause bug in test code.

I'm not sure the best practices for fuzz testing 🥲. For now I could only come up an idea that maybe in the future after we split tests into multiple files, we could separate fuzzing tests from business logic tests to keep them as simple as possible.

contracts/test/LONStaking.t.sol
pilagod: Super powerful! 🤩

foundry.toml
NIC619: Only valid runs

contracts/test/LONStaking.t.sol
NIC619: I rename all of them to `staker` for consistency.

contracts/test/LONStaking.t.sol
NIC619: I ended up moving the `mint` outside out of `_stake` and make it explicit in the tests because only fuzzing tests will need to mint so I think maybe it's not worthy to apply mint logic to all stake helper functions.

contracts/test/LONStaking.t.sol
```diff
@@ -433,80 +534,315 @@ contract LONStakingTest is Test {
...
+
+        for (uint256 i = 0; i < stakeAmounts.length; i++) {
+            address staker = address(uint256(fuzzingUserStartAddress) + i);
+            uint256 stakeAmount = stakeAmounts[i];
+            uint256 redeemAmount = redeemAmounts[i];
+            lon.mint(staker, stakeAmount);
+            _stake(staker, stakeAmount);
+            simulateBuyback(buybackAmounts[i]);
+            // Skip if stake did not get any share due to too small stakeAmount and rounding error
+            if (lonStaking.balanceOf(staker) == 0) continue;
```
NIC619, line 801: I realized that if `stakeAmount` is too small and buyback amounts quickly add up, it still could result in staker getting zero share (despite having staked 10m LON first).
For example, stake `1` would still result in getting `1.mul(10000).div(10001) == 0` share

I tried to check `vm.assume(stakeAmount > 10)` but it results in longer test time (fuzzing likely spend much time looking at edge cases).

contracts/test/LONStaking.t.sol
```diff
@@ -248,17 +249,51 @@ contract LONStakingTest is Test {
...
+        stakerXLON.assertChange(int256(expectedStakeAmount));
+        vm.stopPrank();
+    }
+
+    function testStake() public {
+        _stakeAndValidate(user, DEFAULT_STAKE_AMOUNT);
+    }
+
+    function testFuzz_Stake(uint256 stakeAmount) public {
+        vm.assume(stakeAmount > 0);
```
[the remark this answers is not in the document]
charlesjhongc: I think this check should be done in the staking contract instead. Filtering inputs will narrow the possible scenario of fuzzing.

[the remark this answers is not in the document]
charlesjhongc: Same for `stakeAmount <= lon.cap()` and `stakeAmount.add(lon.totalSupply()) <= lon.cap()`

contracts/test/LONStaking.t.sol
```diff
@@ -248,17 +249,51 @@ contract LONStakingTest is Test {
...
+        _stakeAndValidate(user, stakeAmount);
+    }
+
+    function testFuzz_StakeMultiple(uint80[16] memory stakeAmounts) public {
+        // LON cap lies between 2**87 and 2**88, setting upper bound of stake amount to 2**80 - 1 so stake amount will not exceed cap
+        uint256 totalLONAmount = lon.totalSupply();
+        for (uint256 i = 0; i < stakeAmounts.length; i++) {
+            vm.assume(stakeAmounts[i] > 0);
+            totalLONAmount = totalLONAmount.add(stakeAmounts[i]);
+            vm.assume(totalLONAmount <= lon.cap());
```
charlesjhongc, line 301: Not sure the purpose of this check?

NIC619:
> I think this check should be done in the staking contract instead. Filtering inputs will narrow the possible scenario of fuzzing.

It's already checked in the staking contract so that's why we need to filter it out. We want fuzzing to find corner cases using valid inputs and then we can fix the cases and update the contracts. Have fuzzing testing with with an input that we already know will fail, i.e. invalid inputs, does not help. It just tells you that the input will cause error which is already what we are expecting.

Same applies to checking that total minted LON amount will not exceed LON cap. We already know minting more than LON cap will just fail so we are skipping the inputs that will result in minting more than LON cap.

charlesjhongc:
Seems like foundry doesn't support property assertion so the invariant is `the call should succeed`. Then the filtered input make sense under this context. This is not very ideal for fuzzing but can be a temporary solution till it can be done using better tools.

charlesjhongc (approved):

pilagod (approved):

NIC619:
For the fuzzing tests where we need to build custom scenario instead of random combination of calls, I agree it's not ideal. But for filtering out inputs, it makes perfect sense to me - it's the same as if you don't filter out `stakeAmount == 0` and instead add a condition check `if (stakeAmount == 0) { return; } else { stake(); }` but in this case you simply wasted a fuzz run because you know your function/contract is just not going to be executed.

charlesjhongc:
If the invariant is not `call must succeed` then the input should not be filtered. The invariant should stands for any input and any call sequence. For example `lon.totalSupply() should never exceed lon.cap()` can be an invariant. This stays true even if `stakeAmount` is zero and the transaction is reverted. Filtering out invalid input requires the assumption and prior 
 knowledge of the definition of valid input which is in contradiction to the purpose of fuzzing test.

PR #43, opened 2022-06-23
Adjust init function for new network

charlesjhongc:
~According to Tokenlon contract version control policy, remove version stored in contract. Protect `initialize` function being called twice using `Initializable` contract from OZ. Set operator in `initialize` function.~

Adjust init function for new network deployment. May change design in the future.

### DIFF
--- a/contracts/test/UserProxy.t.sol
+++ b/contracts/test/UserProxy.t.sol
@@ -54,9 +54,9 @@ contract UserProxyTest is Test {
      *********************************/
 
     function testCannotInitializeToZeroAddress() public {
-        UserProxy us = new UserProxy();
+        UserProxy up = new UserProxy();
         vm.expectRevert("UserProxy: operator can not be zero address");
-        us.initialize(address(0));
+        up.initialize(address(0));
     }
 
     function testCannotInitializeAgain() public {
--- a/contracts/test/UserProxy.t.sol
+++ b/contracts/test/UserProxy.t.sol
@@ -19,21 +19,8 @@ contract UserProxyTest is Test {
         // Deploy
         userProxy = new UserProxy();
         strategy = new MockStrategy();
-        // Setup
         // Set this contract as operator
-        // prettier-ignore
-        vm.store(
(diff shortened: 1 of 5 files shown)


NIC619:
Ugly but simple and clear approach:
```solidity
function initialize(xxx) external {
        if (chainID = 1) { // Mainnet
               ...
        } else if (chainID = 42161) { // Arbitrum
               ...
        }
}
```

pilagod:
Another thought is to make initialize idempotent, like

```solidity
function initialize(address _operator) public {
    if (operator == address(0)) {
        operator = _operator
    }
}
```

Instead of constraining initialize can only be called once, we can ensure that it can only take effect once, no matter how many calls.

By this method, we can highlight all required variables that the contract must to have to work in initialize. And new required variable will just append to it like

```solidity
function initialize(address _operator, address _emergencyRecipient) public {
    if (operator == address(0)) {
        operator = _operator
    }
    if (emergencyRecipient == address(0)) {
        emergencyRecipient = _emergencyRecipient
    }
}
```

This can make our contract resilient enough to deploy at any time, any network.

charlesjhongc:
Force pushed to squash commits.

contracts/test/UserProxy.t.sol
```diff
@@ -67,15 +54,19 @@ contract UserProxyTest is Test {
      *********************************/
 
     function testCannotInitializeToZeroAddress() public {
-        vm.expectRevert("UserProxy: _limitOrderAddr should not be 0");
-        userProxy.initialize(address(0));
+        UserProxy us = new UserProxy();
```
pilagod, line 57: nit: Use `up` for consistency

pilagod (approved):

contracts/test/UserProxy.t.sol
charlesjhongc: Forgot this one. Fixed in cc8abb4.

NIC619 (approved):

pilagod (approved):

PR #45, opened 2022-06-27
Redefine Foundry test for AllowanceTarget contract

oneleo:
In order to make the AllowanceTarget contract test files more readable, change these test files into a hierarchical directory.

These test contracts conform to the following specifications:
1) One file, one contract.
2) Test contracts are named with upper camel case.
3) Test functions are named with lower camel case.
4) Minimal import principle: do not import unused libraries.

Please use the following command to test:
```bash
forge test -vvv --force --match-path 'contracts/test/AllowanceTarget/*.t.sol'
```

The following is the test execution result:
```
Running 2 tests for contracts/test/AllowanceTarget/Constructor.t.sol:TestAllowanceTargetConstructor
[PASS] testCannotDeployByZeroAddress() (gas: 41031)
[PASS] testConstructor() (gas: 7650)
Test result: ok. 2 passed; 0 failed; finished in 1.20ms

Running 3 tests for contracts/test/AllowanceTarget/CompleteSetSpender.t.sol:TestAllowanceTargetCompleteSetSpender
[PASS] testCannotCompleteBeforeSet() (gas: 10620)
[PASS] testCannotCompleteTooEarly() (gas: 60817)
[PASS] testCompleteSetSpender() (gas: 48614)
Test result: ok. 3 passed; 0 failed; finished in 1.95ms

Running 2 tests for contracts/test/AllowanceTarget/Teardown.t.sol:TestAllowanceTargetTeardown
[PASS] testCannotTeardownByNotSpender() (gas: 13059)
[PASS] testTeardown() (gas: 14054)
Test result: ok. 2 passed; 0 failed; finished in 1.87ms

Running 2 tests for contracts/test/AllowanceTarget/ExecuteCall.t.sol:TestAllowanceTargetExecuteCall
[PASS] testCannotExecuteByNotSpender() (gas: 16015)
[PASS] testExecuteCall() (gas: 832810)
Test result: ok. 2 passed; 0 failed; finished in 2.11ms

Running 4 tests for contracts/test/AllowanceTarget/SetSpenderWithTimelock.t.sol:TestAllowanceTargetSetSpenderWithTimeock
[PASS] testCannotSetByInvalidAddress() (gas: 15503)
[PASS] testCannotSetByRandomEOA() (gas: 15318)
[PASS] testCannotSetInProgress() (gas: 141333)
[PASS] testSetSpenderWithTimeock() (gas: 58853)
Test result: ok. 4 passed; 0 failed; finished in 2.05ms
```

Thanks a lot!

### DIFF
--- a/contracts/test/AllowanceTarget/CompleteSetSpender.t.sol
+++ b/contracts/test/AllowanceTarget/CompleteSetSpender.t.sol
@@ -4,12 +4,12 @@ pragma solidity 0.7.6;
 import "contracts-test/AllowanceTarget/Setup.t.sol";
 
 contract TestAllowanceTargetCompleteSetSpender is TestAllowanceTarget {
-    function testCannotCompleteBeforeSet() public {
+    function testCannotCompleteSetSpenderBeforeSet() public {
         vm.expectRevert("AllowanceTarget: no pending SetSpender");
         allowanceTarget.completeSetSpender();
     }
 
-    function testCannotCompleteTooEarly() public {
+    function testCannotCompleteSetSpenderTooEarly() public {
         allowanceTarget.setSpenderWithTimelock(newSpender);
         vm.expectRevert("AllowanceTarget: time lock not expired yet");
         allowanceTarget.completeSetSpender();
--- a/contracts/test/AllowanceTarget/CompleteSetSpender.t.sol
+++ b/contracts/test/AllowanceTarget/CompleteSetSpender.t.sol
@@ -4,18 +4,18 @@ pragma solidity 0.7.6;
 import "contracts-test/AllowanceTarget/Setup.t.sol";
 
 contract TestAllowanceTargetCompleteSetSpender is TestAllowanceTarget {
-    function testCannotCompleteSetSpenderBeforeSet() public {
+    function testCannotSetBeforeSetNew() public {
         vm.expectRevert("AllowanceTarget: no pending SetSpender");
         allowanceTarget.completeSetSpender();
     }
 
-    function testCannotCompleteSetSpenderTooEarly() public {
+    function testCannotSetTooEarly() public {
         allowanceTarget.setSpenderWithTimelock(newSpender);
         vm.expectRevert("AllowanceTarget: time lock not expired yet");
         allowanceTarget.completeSetSpender();
     }
 
-    // usually case
+    // normal case
     function testCompleteSetSpender() public {
         allowanceTarget.setSpenderWithTimelock(newSpender);
         // fast forward
--- a/contracts/test/AllowanceTarget/CompleteSetSpender.t.sol
+++ b/contracts/test/AllowanceTarget/CompleteSetSpender.t.sol
@@ -4,18 +4,17 @@ pragma solidity 0.7.6;
 import "contracts-test/AllowanceTarget/Setup.t.sol";
 
 contract TestAllowanceTargetCompleteSetSpender is TestAllowanceTarget {
-    function testCannotSetBeforeSetNew() public {
+    function testCannotCompleteBeforeSet() public {
         vm.expectRevert("AllowanceTarget: no pending SetSpender");
         allowanceTarget.completeSetSpender();
     }
 
-    function testCannotSetTooEarly() public {
+    function testCannotCompleteTooEarly() public {
         allowanceTarget.setSpenderWithTimelock(newSpender);
         vm.expectRevert("AllowanceTarget: time lock not expired yet");
         allowanceTarget.completeSetSpender();
     }
 
-    // normal case
     function testCompleteSetSpender() public {
         allowanceTarget.setSpenderWithTimelock(newSpender);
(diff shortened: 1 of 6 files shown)


charlesjhongc (commented):
`usually` case -> normal case or usual case

contracts/test/AllowanceTarget/Constructor.t.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.7.6;
+
+import "contracts-test/AllowanceTarget/Setup.t.sol";
+
+contract TestAllowanceTargetConstructor is TestAllowanceTarget {
+    function testCannotDeployByZeroAddress() public {
```
NIC619, line 7: I think we need to differentiate between test case names event if the test cases are separated in different files. For example `testCannotDeployByZeroAddress` and `testConstructor` could exist in most of the contracts' test cases, so if you do `forge test --match-test testConstructor` it will run all `testConstructor ` test cases in those contracts' test cases.

contracts/test/AllowanceTarget/ExecuteCall.t.sol
```diff
@@ -0,0 +1,30 @@
...
+
+import "contracts-test/AllowanceTarget/Setup.t.sol";
+import "contracts-test/utils/BalanceSnapshot.sol";
+import "contracts-test/mocks/MockERC20.sol";
+
+contract TestAllowanceTargetExecuteCall is TestAllowanceTarget {
+    // include Snapshot struct from BalanceSnapshot
+    using BalanceSnapshot for BalanceSnapshot.Snapshot;
+
+    function testCannotExecuteByNotSpender() public {
```
NIC619, line 12: Same for test name like this, it would better to include the whole function name in there: `testCannotExecuteCallByNotSpender`

contracts/test/AllowanceTarget/Constructor.t.sol
charlesjhongc: But this way the case will still have lengthy naming. This is a workaround so not the most ideal way and I'm ok with sacrificing `--match-test` for short case naming.

contracts/test/AllowanceTarget/ExecuteCall.t.sol
oneleo: No problem! I've filled in the full test function name.

contracts/test/AllowanceTarget/Constructor.t.sol
charlesjhongc: If the main goal is to reduce lines in a single file, not the length of the name, then it make sense.

pilagod: I think this is not a problem since we can use `--match-test` together with `--match-contract` or `--match-path`. 

I think lengthy case name is one of the pain points we try to solve by slice tests into multiple files, right?

oneleo: Thanks for suggestions! Maybe we should discuss the choice of long and short test func names.

contracts/test/AllowanceTarget/CompleteSetSpender.t.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.7.6;
+
+import "contracts-test/AllowanceTarget/Setup.t.sol";
+
+contract TestAllowanceTargetCompleteSetSpender is TestAllowanceTarget {
+    function testCannotSetBeforeSetNew() public {
```
pilagod, line 7: Would it be more smooth to name by `complete` instead of `set`? For example, `testCannotCompleteBeforeSet `

contracts/test/AllowanceTarget/CompleteSetSpender.t.sol
```diff
@@ -0,0 +1,26 @@
...
+        allowanceTarget.completeSetSpender();
+    }
+
+    function testCannotSetTooEarly() public {
+        allowanceTarget.setSpenderWithTimelock(newSpender);
+        vm.expectRevert("AllowanceTarget: time lock not expired yet");
+        allowanceTarget.completeSetSpender();
+    }
+
+    // normal case
```
pilagod, line 18: I think this comment doesn't transmit too much info. Maybe we could just remove it

oneleo: OK! Is it only in this CompleteSetSpender.t.sol file?
Or should I remove all "normal case" comments in all *.t.sol contract files?

changwu-tw (approved):

changwu-tw:
Having an execution command in PR is good for the reviewer to run test. Keep it up.

pilagod (approved):

charlesjhongc (approved):
Then remove original test `AllowanceTarget.t.sol`. Can be done in another PR.

PR #55, opened 2022-07-08
Remove AMM subsidy

charlesjhongc:
## Adjust `Swapped` event fields
- removed `receivedAmount`
- removed `subsidyFactor`
- use shared event definition for both `AMMWrapper` and `AMMWrapperWithPath`


contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
```diff
@@ -1,62 +1,62 @@
...
+import "contracts/AMMWrapper.sol";
 import "contracts-test/forkMainnet/AMMWrapper/Setup.t.sol";
 import "contracts-test/utils/BalanceSnapshot.sol";
 
 contract TestAMMWrapperCollectFee is TestAMMWrapper {
     using BalanceSnapshot for BalanceSnapshot.Snapshot;
 
-    function testCollectFeeIfReceivedSameAsMinOut() public {
-        uint256 feeFactor = 5;
+    function testFailedIfMakerAssetAmountNotDeductFee() public {
```
NIC619, line 11: IMO `MakerAssetAmountNotDeductFee` is not so straightforward. From contract's point of view, the only reason it fails is because the minimum amount is not high enough after the it adds fee on top of `makerAssetAmount`. It has no idea about whether fee is subtracted from `makerAssetAmount` or not.

Maybe `testCannotCollectFeeIfReceivedLessThanMinOutPlusFee`?

contracts/AMMWrapper.sol
```diff
@@ -219,29 +222,19 @@ contract AMMWrapper is IAMMWrapper, ReentrancyGuard, BaseLibEIP712, SignatureVal
...
-            order.makerAssetAmount,
-            order.receiverAddr,
-            txMetaData.settleAmount,
-            txMetaData.receivedAmount,
-            txMetaData.feeFactor,
-            txMetaData.subsidyFactor
-        );
+            // Settle
+            // Calculate fee using actually received from swap
+            uint256 actualFee = txMetaData.receivedAmount.mul(txMetaData.feeFactor).div(LibConstant.BPS_MAX);
```
NIC619, line 232: The fee will be different from the one computed above (in line 227). If this is intended (i.e., we can collect more fee if received more), we should also update the spec.
Otherwise we should probably store the fee in a local variable.

contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
```diff
@@ -1,62 +1,62 @@
...
         bytes memory sig = _signTrade(userPrivateKey, order);
         bytes memory payload = _genTradePayload(order, feeFactor, sig);
 
         BalanceSnapshot.Snapshot memory ammWrapperMakerAsset = BalanceSnapshot.take(address(ammWrapper), order.makerAssetAddr);
 
         userProxy.toAMM(payload);
 
-        ammWrapperMakerAsset.assertChangeGt(int256(0));
+        uint256 actualFee = (expectedOutAmount * ammWrapper.defaultFeeFactor()) / BPS_MAX;
+        ammWrapperMakerAsset.assertChange(int256(actualFee));
```
NIC619, line 60: I think we can replace the balance diff check with `feeFactor` param in `Swapped` event?

NIC619:
> - removed `receivedAmount`

Just double checking to make sure if the backend team is aware of the change.

contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
```diff
@@ -1,62 +1,62 @@
...
+        uint256 feeFactor = 100;
         AMMLibEIP712.Order memory order = DEFAULT_ORDER;
         uint256 expectedOutAmount = ammQuoter.getMakerOutAmount(order.makerAddr, order.takerAssetAddr, order.makerAssetAddr, order.takerAssetAmount);
         order.makerAssetAmount = expectedOutAmount;
         bytes memory sig = _signTrade(userPrivateKey, order);
         bytes memory payload = _genTradePayload(order, feeFactor, sig);
 
-        BalanceSnapshot.Snapshot memory ammWrapperMakerAsset = BalanceSnapshot.take(address(ammWrapper), order.makerAssetAddr);
-
+        vm.expectRevert("AMMWrapper: not the operator");
```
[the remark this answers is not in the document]
charlesjhongc: I forgot to change revert message but then I wondered why it got passed. After debugging for almost an hour, the answer is that the name of the test match reserved keywords `testFail....` of foundry. So foundry check whether it's reverted and seems `vm.expectRevert` has no effect in this situation. Fixed in 72c773d.

contracts/AMMWrapper.sol
charlesjhongc: This is intended and aligned with the spec. The reason of this is to prevent user bringing too small `makerAssetAmount` although it's not likely.

contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
charlesjhongc: Is there a way to get param in the emitted event?

contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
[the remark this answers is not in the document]
NIC619: I see... this definitely worth being noted.

contracts/AMMWrapper.sol
NIC619: What I'm mean is that the `actualFee` computed in `_swap` and `_settle` would be different while in the spec it's like this:
```solidity
uint256 fee = _order.makerAssetAmount.mul(feeFactor).div(BPS_MAX);
// AMM should output with minAmount = makerAssetAmount + fee 
uint256 minAmount = _order.makerAssetAmount.add(fee);
uint256 receivedAmount = _swap(_order, minAmount);

// Transfer (receivedAmount - fee) to user
uint256 settleAmount = receivedAmount.sub(fee);
token.safeTransfer(_order.receiverAddr, settleAmount);
```

so we should probably align them.

contracts/AMMWrapper.sol
```diff
@@ -365,68 +358,23 @@ contract AMMWrapper is IAMMWrapper, ReentrancyGuard, BaseLibEIP712, SignatureVal
 
     /**
      * @dev internal function of `trade`.
-     * It collects fee from the trade or compensates the trade based on the actual amount swapped.
+     * It collects fee from the trade and xxxxxxxxxxxxxxxxxxxx.
```
pilagod, line 361: Does this forget to replace 😆

contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
pilagod: I think balance check is intuitive for me in this scenario.

contracts/test/forkMainnet/AMMWrapperWithPath/TradeUniswapV3.t.sol
```diff
@@ -187,41 +201,47 @@ contract TestAMMWrapperWithPathTradeUniswapV3 is TestAMMWrapperWithPath {
...
-            expectedOutAmount, // settleAmount: no fee so settled amount is the same as received amount
-            expectedOutAmount, // receivedAmount
-            uint16(feeFactor), // Fee factor: 0
-            uint16(0) // Subsidy factor: 0
-        );
-        emit Swapped(txMetaData, order);
+        bytes memory makerSpecificData = _encodeUniswapMultiPoolData(MULTI_POOL_SWAP_TYPE, DEFAULT_MULTI_HOP_PATH, DEFAULT_MULTI_HOP_POOL_FEES);
+        bytes memory payload = _genTradePayload(order, feeFactor, _signTrade(userPrivateKey, order), makerSpecificData, DEFAULT_MULTI_HOP_PATH);
+
+        {
```
pilagod, line 219: Is this block to avoid stack too deep?

contracts/AMMWrapper.sol
charlesjhongc: My bad. I think the spec should be updated if we all accept the logic in this PR.

contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
charlesjhongc: Yeah but I can add event check as well. Will do.

contracts/AMMWrapper.sol
NIC619: I support this one too. Let's update the spec.

pilagod: Gogogo

contracts/test/forkMainnet/AMMWrapperWithPath/TradeUniswapV3.t.sol
```diff
@@ -187,41 +201,47 @@ contract TestAMMWrapperWithPathTradeUniswapV3 is TestAMMWrapperWithPath {
...
         userProxy.toAMM(payload);
 
         userTakerAsset.assertChange(-int256(order.takerAssetAmount));
         userMakerAsset.assertChangeGt(int256(order.makerAssetAmount));
     }
 
     function testEmitSwappedEvent() public {
+        // this one was make compact in order to avoid stack too deep
+        // focusing on event fileds in this case
         uint256 feeFactor = 0;
```
changwu-tw, line 214: Should we include the case while the sender is the user himself instead of relayer, so we can check the feeFactor will be set to the defaultFeeFactor ?

contracts/AMMWrapper.sol
changwu-tw: No wonder I'm a little bit confused. If so, it might be involved with changing FE logic. For the app, it needs to handle two different logics from old and new contract?

charlesjhongc: I'll confirm with FE team.

contracts/AMMWrapper.sol
charlesjhongc: 🤦‍♂️ Fixed in 97f4251.

contracts/test/forkMainnet/AMMWrapperWithPath/TradeUniswapV3.t.sol
charlesjhongc: It's already covered in `CollectFee.t.sol`

contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
charlesjhongc: Fixed in c276658.

pilagod (dismissed):

contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
```diff
@@ -1,62 +1,77 @@
...
+            order.userAddr,
+            false, // not relayed
+            order.takerAssetAddr,
+            order.takerAssetAmount,
+            order.makerAddr,
+            order.makerAssetAddr,
+            order.makerAssetAmount,
+            order.receiverAddr,
+            expectedOutAmount - actualFee, // default fee factor will be applied
+            ammWrapper.defaultFeeFactor()
```
NIC619, line 71:
```suggestion
            expectedOutAmount - actualFee,
            ammWrapper.defaultFeeFactor() // default fee factor will be applied
```

NIC619 (approved):

pilagod (approved):

PR #57, opened 2022-07-13
Verify sig in MockERC1271Wallet

NIC619:
- Verify sig in `MockERC1271Wallet` so we can know at least if our signing utils in tests are correct

### DIFF
--- a/contracts/test/mocks/MockERC1271Wallet.sol
+++ b/contracts/test/mocks/MockERC1271Wallet.sol
@@ -2,17 +2,20 @@ pragma solidity 0.7.6;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+import "contracts/utils/LibBytes.sol";
 import "../../interfaces/ISetAllowance.sol";
 import "../../interfaces/IERC1271Wallet.sol";
 
 contract MockERC1271Wallet is ISetAllowance, IERC1271Wallet {
     using SafeERC20 for IERC20;
+    using LibBytes for bytes;
+
     // bytes4(keccak256("isValidSignature(bytes,bytes)"))
     bytes4 internal constant ERC1271_MAGICVALUE = 0x20c13b0b;
-
     // bytes4(keccak256("isValidSignature(bytes32,bytes)"))
     bytes4 internal constant ERC1271_MAGICVALUE_BYTES32 = 0x1626ba7e;
     uint256 private constant MAX_UINT = 2**256 - 1;
+
     address public operator;
 
     modifier onlyOperator() {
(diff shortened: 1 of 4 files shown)


NIC619:
So it turns out that ERC1271 wallet can't be used for PMM because 0x's signature does not match the wallet standard...

pilagod (approved):
Just curious about that is there any standard ruling the r, s, v order? 

I see EIP1271 doesn't mention any signature details. It seems to allow everyone to implement specific wallet, like (r, v, s), (s, r, v) 🤣

contracts/test/mocks/MockERC1271Wallet.sol
```diff
@@ -37,10 +40,19 @@ contract MockERC1271Wallet is ISetAllowance, IERC1271Wallet {
...
+        require(operator == _ecrecover(keccak256(_data), _signature), "MockERC1271Wallet: invalid signature");
         return ERC1271_MAGICVALUE;
     }
 
     function isValidSignature(bytes32 _hash, bytes calldata _signature) external view override returns (bytes4 magicValue) {
+        require(operator == _ecrecover(_hash, _signature), "MockERC1271Wallet: invalid signature");
         return ERC1271_MAGICVALUE_BYTES32;
     }
+
+    function _ecrecover(bytes32 hash, bytes memory signature) internal pure returns (address) {
```
charlesjhongc, line 52: How about using OZ lib for this?
https://docs.openzeppelin.com/contracts/3.x/api/cryptography#ECDSA-recover-bytes32-bytes-

NIC619:
> Just curious about that is there any standard ruling the r, s, v order?

Yeah there does not seem to be a standard. Can only presume people will follow popular library like OZ's.🧐

contracts/test/mocks/MockERC1271Wallet.sol
NIC619: Updated in c369e28

contracts/test/mocks/MockERC1271Wallet.sol
```diff
@@ -49,10 +50,10 @@ contract MockERC1271Wallet is ISetAllowance, IERC1271Wallet {
         return ERC1271_MAGICVALUE_BYTES32;
     }
 
-    function _ecrecover(bytes32 hash, bytes memory signature) internal pure returns (address) {
+    function _ecrecover(bytes32 _hash, bytes memory signature) internal pure returns (address) {
```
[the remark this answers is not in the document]
NIC619: Fixed in 7b38090

pilagod (dismissed):

contracts/test/mocks/MockERC1271Wallet.sol
NIC619: Replace `_ecrecover` with `ECDSA.recover` in 59fc2da

charlesjhongc (approved):

pilagod (approved):

PR #58, opened 2022-07-15
veLON contract with early withdrawal penalty

zeqing-guo:
1. Forked from [veMULTI](https://github.com/anyswap/veMULTI/blob/main/contracts/ve.sol)
2. Supported early withdrawal penalty
3. Add some foundry test cases: 
    1. initializing the project
    2. testing staking
    3. testing early withdrawal
4. Improve the coding style


NIC619:
It would be great if commit 33261c3 can be split into two: first copy ve.sol and second apply changes on top. Then we can review second commit and see the new changes applied to ve.sol.

charlesjhongc:
After discussion, I'll refactor the implementation in another PR. Keep this one opened for now so new feature and test can be added without being blocked.

charlesjhongc:
Let's keep this PR opened as a reference until refactored implementation get merged.

PR #59, opened 2022-07-21
add basic functions of veLON

charlesjhongc:
Add basic functions like `deposit` and `withdraw` of veLON with `_checkpoint()` calculation.

### DIFF
--- a/contracts/veLON.sol
+++ b/contracts/veLON.sol
@@ -163,6 +163,25 @@ contract veLON is IveLON, ERC721, Ownable, ReentrancyGuard {
         emit Withdraw(msg.sender, expired, _tokenId, amount, penalty, block.timestamp);
     }
 
+    /// @notice Merge two locking NFTs as one
+    /// @param _from NFT that holds lock and to be burned
+    /// @param _to NFT that holds lock and to be updated
+    function merge(uint256 _from, uint256 _to) external override {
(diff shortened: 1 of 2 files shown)


contracts/veLON.sol
```diff
@@ -0,0 +1,347 @@
...
+    uint256 public tokenSupply;
+    uint256 public epoch;
+    uint256 public maxLockDuration = 365 days;
+    uint256 public earlyWithdrawPenaltyRate = 3000;
+
+    mapping(uint256 => Point) public poolPointHistory; // epoch -> unsignd point
+    mapping(uint256 => Point[1000000000]) public userPointHistory; // user -> Point[user_epoch]
+    mapping(uint256 => LockedBalance) public locked; // tokenId -> locked balance
+    mapping(uint256 => uint256) public userPointEpoch; // tokenId -> epoch
+    mapping(uint256 => int256) public slope_changes; // time -> signed slope change
```
zeqing-guo, line 31: We might change this to `slopeChange`

charlesjhongc: Actually I left it with underscore because I was thinking of updating couple namings in this contract. But let's fix the case first. Fixed in 9a24720.

charlesjhongc:
Force pushed to discard informal commit message.

PR #60, opened 2022-07-26
Add feeCollector to AMM and transfer fee away

NIC619:
- Add `feeCollector`
- Transfer fee to `feeCollector`


contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
```diff
@@ -34,12 +34,12 @@ contract TestAMMWrapperCollectFee is TestAMMWrapper {
         bytes memory sig = _signTrade(userPrivateKey, order);
         bytes memory payload = _genTradePayload(order, feeFactor, sig);
 
-        BalanceSnapshot.Snapshot memory ammWrapperMakerAsset = BalanceSnapshot.take(address(ammWrapper), order.makerAssetAddr);
+        BalanceSnapshot.Snapshot memory feeCollectorMakerAsset = BalanceSnapshot.take(address(feeCollector), order.makerAssetAddr);
```
pilagod, line 37: Maybe we could remove `address()` wrap since `feeCollector` is already an address value

contracts/test/forkMainnet/AMMWrapper/SetFeeCollector.t.sol
```diff
@@ -0,0 +1,17 @@
...
+
+contract TestAMMWrapperSetFeeCollector is TestAMMWrapper {
+    function testCannotSetFeeCollectorByNotOperator() public {
+        vm.prank(user);
+        vm.expectRevert("AMMWrapper: not the operator");
+        ammWrapper.setFeeCollector(user);
+    }
+
+    function testSetFeeCollector() public {
+        ammWrapper.setFeeCollector(user);
```
charlesjhongc, line 14: Check event as well?

contracts/AMMWrapper.sol
```diff
@@ -376,6 +390,20 @@ contract AMMWrapper is IAMMWrapper, ReentrancyGuard, BaseLibEIP712, SignatureVal
             // other ERC20 tokens
             IERC20(_order.makerAssetAddr).safeTransfer(_order.receiverAddr, _settleAmount);
         }
+        // Collect fee
+        if (_feeAmount > 0) {
```
charlesjhongc, line 394: I think we should add ETH/WETH fee collecting test. Can be done in another PR.

contracts/AMMWrapper.sol
```diff
@@ -376,6 +390,20 @@ contract AMMWrapper is IAMMWrapper, ReentrancyGuard, BaseLibEIP712, SignatureVal
...
+            if (_internalTxData.toEth) {
+                // Transfer WETH directly if internal maker asset is WETH
+                if (!_isInternalAssetETH(_internalTxData.makerAssetInternalAddr)) {
+                    weth.transfer(feeCollector, _feeAmount);
+                } else {
+                    payable(feeCollector).transfer(_feeAmount);
+                }
```
pilagod, line 401: Do you think it is a good idea to align the way of collecting fee with settling for receiver? 

It seems that receiver will always get ETH no matter internal asset is ETH or WETH. Maybe we could do the same for fee collector for better fund management?

NIC619: This saves some gas (1 eth transfer instead of 2) for every trade (with WETH as internal asset) and is worth it in my opinion but I can change if we think otherwise.

contracts/AMMWrapper.sol
NIC619: Will do

contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
NIC619: Fixed in d281833

contracts/test/forkMainnet/AMMWrapper/SetFeeCollector.t.sol
NIC619: Added in d281833

contracts/AMMWrapper.sol
NIC619: Added in 47d167a

contracts/AMMWrapper.sol
pilagod: Yeah, I think it's a cost vs operating problem, and there is no absolutely right answer. 😂 

Align the logic would make code simpler and more consistent (could even extract an util function to share the transfer logic for both receiver and fee collector), and fee collector would always receive ETH, easier for him to use fee as gas to send other tx in the future.

Actually I'm fine with both implementations. 💪 Just raise a different point of view here. Maybe we could refer to @charlesjhongc opinions.

contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
```diff
@@ -42,6 +42,53 @@ contract TestAMMWrapperCollectFee is TestAMMWrapper {
         feeCollectorMakerAsset.assertChange(int256(fee));
     }
 
+    function testCollectFeeWithWETH() public {
+        uint256 feeFactor = 100;
+        AMMLibEIP712.Order memory order = DEFAULT_ORDER;
+        order.makerAssetAddr = ETH_ADDRESS;
+        order.makerAssetAmount = 0.001 ether;
```
pilagod, line 49: Maker asset amount would be overridden at later, maybe we could skip this assignment to avoid confusion?

contracts/AMMWrapper.sol
charlesjhongc: From the operational perspective, buyback only deal with WETH currently so I think fee should stay wrapped when possible. It will cause few inconsistencies in code but save some gas for user/relayer every time so the trade-off is worthy to me.

charlesjhongc (dismissed):
LGTM

contracts/test/forkMainnet/AMMWrapper/CollectFee.t.sol
NIC619: Fixed in e0efdc2

pilagod (approved):

charlesjhongc (approved):

PR #63, opened 2022-07-28
Add ve test after refactor

eric6311:
add basic tests for veLon after refactoring
- test setup
- test deposit (including CreateLock, DepositFor, ExtendLock, Merge)
- test withdraw
- test transfer 

The test files are decomposed into hierarchical files, use command below to run the test: 
`forge test -vv --match-path "contracts/test/veLON/*.t.sol"`


contracts/test/veLON.t.sol
```diff
@@ -0,0 +1,265 @@
...
+        // uint256 numberOfNftBefore = veLon.totalNFTSupply();
+        // uint256 totalVePowerBefore = veLon.totalSupply();
+        vm.startPrank(staker);
+        if (lon.allowance(staker, address(veLon)) == 0) {
+            lon.approve(address(veLon), type(uint256).max);
+        }
+        uint256 tokenId = veLon.createLock(stakeAmount, lockDuration);
+        stakerLon.assertChange(-int256(stakeAmount));
+        lockedLon.assertChange(int256(stakeAmount));
+        // assertEq(veLon.totalNFTSupply(), numberOfNftBefore + 1);
```
zeqing-guo, line 99: I think we need to validate the NFT supply and NFT's voting power.

contracts/test/veLON.t.sol
```diff
@@ -0,0 +1,265 @@
...
+        uint256 tokenId = _stakeAndValidate(user, DEFAULT_STAKE_AMOUNT, 2 weeks);
+
+        stakerLon = BalanceSnapshot.take(user, address(lon));
+        lockedLon = BalanceSnapshot.take(address(veLon), address(lon));
+
+        //pretend 1 week has passed and the lock not expired
+        vm.warp(block.timestamp + 1 weeks);
+        vm.prank(user);
+        veLon.withdrawEarly(tokenId);
+        uint256 balanceChange = DEFAULT_STAKE_AMOUNT.mul(earlyWithdrawPenaltyRate).div(PENALTY_RATE_PRECISION);
```
zeqing-guo, line 228: The `earlyWithdrawPenaltyRate` is mutable and we may change it to another number someday, which would make the test fail. Thus, it is better to get the value from the contract directly.

contracts/test/veLON.t.sol
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    function testCannoExtendLockExpired() public {
+        uint256 lockTime = 1 weeks;
+        uint256 tokenId = _stakeAndValidate(user, DEFAULT_STAKE_AMOUNT, lockTime);
+
+        //pretend 1 week has passed and the lock expired
+        vm.warp(block.timestamp + 1 weeks);
+        vm.prank(user);
+        vm.expectRevert("Lock expired");
```
zeqing-guo, line 177: We also need to test `"Not approved or owner" revert` for `extendLock` function.

contracts/test/veLON.t.sol
eric6311: The voting power for single NFT is not visible in current veLON.sol, so I left the change in the future

eric6311:
`merge to add-veLON-implementation`, so close this request

PR #65, opened 2022-07-29
Add `setMaxLockDuration` function to `add-veLON-impplementation`

zeqing-guo:
Add `setMaxLockDuration` function to `add-veLON-impplementation` branch.

Given the correctness and fairness, the code guarantees that if `old_power_A > old_power_B`, then after updating the `maxLockDuration` we have `new_power_A >= new_power_B`.

Code details are as follows:

- if NFT's `lock.end > maxEndTime`, we set the `lock.end = maxEndTime`. Then we update NFT's slope and bias
- if NFT's `lock.end <= maxEndTime`, we update NFT's slope and bias directly

After updating every NFT's end, slope, and bias, we will update the global slope and bias.

### DIFF
--- a/contracts/interfaces/IveLON.sol
+++ b/contracts/interfaces/IveLON.sol
@@ -14,13 +14,13 @@ interface IveLON is IERC721, IERC721Metadata {
     }
 
     struct LockedBalance {
-        int128 amount; // FIXME use uint256?
+        uint256 amount;
         uint256 end;
     }
 
     struct Point {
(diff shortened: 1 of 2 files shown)


charlesjhongc:
No particular issue at the first glance. However this function may consume lots of gas and perhaps exceed block limit? Need some test for gas consumption.

charlesjhongc:
This one can be closed since we decided to use fixed `maxDuration` instead.

PR #66, opened 2022-07-29
Add ve test after refactor

eric6311:
add basic tests for veLon after refactoring

test setup
test deposit (including CreateLock, DepositFor, ExtendLock, Merge)
test withdraw
test transfer

The test files are decomposed into hierarchical files, use command below to run the test:
`forge test -vv --match-path "contracts/test/veLON/*.t.sol"`


contracts/test/veLON/Setup.t.sol
```diff
@@ -0,0 +1,113 @@
...
+    Lon lon = new Lon(address(this), address(this));
+    veLON veLon;
+
+    uint256 constant DEFAULT_STAKE_AMOUNT = 1e18;
+    uint256 constant MAX_LOCK_TIME = 365 days;
+    uint256 public constant PENALTY_RATE_PRECISION = 10000;
+
+    uint256 public earlyWithdrawPenaltyRate;
+
+    //record the balnce of Lon and VeLon in VM
```
charlesjhongc, line 28: To be consistent, leave one space between `//` and comments.

contracts/test/veLON/Setup.t.sol
```diff
@@ -0,0 +1,113 @@
...
+
+    function setUp() public {
+        // Setup
+        veLon = new veLON(address(lon));
+
+        // Deal 100 ETH to user
+        deal(user, 100 ether);
+        // Mint LON to user
+        lon.mint(user, 100 * 1e18);
+        // User approve LONStaking
```
charlesjhongc, line 40: LONStaking ?

charlesjhongc: Use `veLon` is better. LONStaking is another contract.

contracts/test/veLON/Setup.t.sol
```diff
@@ -0,0 +1,113 @@
...
+        vm.prank(user);
+        lon.approve(address(veLon), type(uint256).max);
+
+        // Deal 100 ETH to otherUSer
+        deal(other, 100 ether);
+        // Mint LON to otherUser
+        lon.mint(other, 100 * 1e18);
+        // User approve LONStaking
```
charlesjhongc, line 48: Same here.

contracts/test/veLON/Setup.t.sol
```diff
@@ -0,0 +1,113 @@
...
+
+        // Deal 100 ETH to user
+        deal(user, 100 ether);
+        // Mint LON to user
+        lon.mint(user, 100 * 1e18);
+        // User approve LONStaking
+        vm.prank(user);
+        lon.approve(address(veLon), type(uint256).max);
+
+        // Deal 100 ETH to otherUSer
```
charlesjhongc, line 44: `otherUser` ?

contracts/test/veLON/Setup.t.sol
```diff
@@ -0,0 +1,113 @@
...
+        deal(other, 100 ether);
+        // Mint LON to otherUser
+        lon.mint(other, 100 * 1e18);
+        // User approve LONStaking
+        vm.prank(other);
+        lon.approve(address(veLon), type(uint256).max);
+
+        //set earlyWithdrawPenaltyRate from veLon
+        earlyWithdrawPenaltyRate = veLon.earlyWithdrawPenaltyRate();
```
charlesjhongc, line 53: How about read this value in case when it's needed? Since not every testing case depends on this.

contracts/test/veLON/Setup.t.sol
```diff
@@ -0,0 +1,113 @@
...
+        assertEq(veLon.maxLockDuration(), MAX_LOCK_TIME);
+        assertEq(veLon.earlyWithdrawPenaltyRate(), 3000);
+    }
+
+    /*********************************
+     *         Test: stake           *
+     *********************************/
+    //compute the power added when staking amount added
+    function _vePowerAdd(uint256 stakeAmount, uint256 lockDuration) internal returns (uint256) {
+        uint256 unlockTime = ((block.timestamp + lockDuration) / 1 weeks) * 1 weeks; // Locktime is rounded down to weeks
```
charlesjhongc, line 80: To be more precise, the rounded is unlock time not lock time.

contracts/test/veLON/Setup.t.sol
```diff
@@ -0,0 +1,113 @@
...
+        uint256 power = (stakeAmount / MAX_LOCK_TIME) * (unlockTime - block.timestamp);
+        return power;
+    }
+
+    function _stakeAndValidate(
```
charlesjhongc, line 85: Perhaps having a basic test for `createLock()` and call `_stakeAndValidate()` inside the test?

contracts/test/veLON/Setup.t.sol
eric6311: Those comments were copied from other  contract. I'l remove them later

contracts/test/veLON/Setup.t.sol
eric6311: Understood

contracts/test/veLON/Setup.t.sol
eric6311: Other user means the second user to interact with. I have added more comments to explain it.

contracts/test/veLON/Setup.t.sol
eric6311: The basic creatLock test is written in `Deposit.t.sol` and it calls ` _stakeAndValidate()` to test it.

charlesjhongc: Oh I see. It make sense this way.

contracts/test/veLON/Setup.t.sol
```diff
@@ -0,0 +1,109 @@
...
+    function testSetupveLON() public {
+        assertEq(veLon.owner(), address(this));
+        assertEq(address(veLon.token()), address(lon));
+        assertEq(veLon.tokenSupply(), 0);
+        assertEq(veLon.maxLockDuration(), MAX_LOCK_TIME);
+        assertEq(veLon.earlyWithdrawPenaltyRate(), 3000);
+    }
+
+    /*********************************
+     *         Test: stake           *
```
charlesjhongc, line 71: Then perhaps this comment should be adjusted like `staking utils` since these are not the real test of staking?

eric6311: I think it would be better to replace it with `staking utils ` too.

charlesjhongc (approved):

PR #68, opened 2022-08-01
Add vBalance and supply view functions

charlesjhongc:
- Add vBalance view function for user
- Add total vBalance view function for pool global
- Add `ownershipChange` to prevent flash NFT
- Override NFT `_transfer` function to set `ownershipChange` first

### DIFF
--- a/contracts/veLON.sol
+++ b/contracts/veLON.sol
@@ -393,8 +393,8 @@ contract veLON is IveLON, ERC721, Ownable, ReentrancyGuard {
 
     // Go over weeks to fill history and calculate what the current pool point is
     function _syncGlobalPoints(Point memory poolLastPoint, uint256 _epoch) private returns (Point memory, uint256) {
-        // block slope = dBlock/dTime
-        // If last point is already recorded in this block, slope=0
+        // block decliningRate = dBlock/dTime
+        // If last point is already recorded in this block, decliningRate=0
         // But that's ok because we know the block in such case
         uint256 blockSlope = 0;
         if (block.timestamp > poolLastPoint.ts) {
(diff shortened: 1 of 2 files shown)


contracts/veLON.sol
```diff
@@ -285,8 +394,8 @@ contract veLON is IveLON, ERC721, Ownable, ReentrancyGuard {
 
     // Go over weeks to fill history and calculate what the current pool point is
     function _syncGlobalPoints(Point memory poolLastPoint, uint256 _epoch) private returns (Point memory, uint256) {
-        // block slope = dBlock/dTime
-        // If last point is already recorded in this block, slope=0
+        // block decliningRate = dBlock/dTime
+        // If last point is already recorded in this block, decliningRate=0
         // But that's ok because we know the block in such case
         uint256 blockSlope = 0;
```
zeqing-guo, line 400: Should we change `blockSlope` to `blockDecliningRate`?

contracts/veLON.sol
```diff
@@ -42,6 +43,114 @@ contract veLON is IveLON, ERC721, Ownable, ReentrancyGuard {
         poolPointHistory[0].ts = block.timestamp;
     }
 
+    function vBalanceOf(uint256 _tokenId) external view override returns (uint256) {
+        return _vBalanceOfAtTime(_tokenId, block.timestamp);
+    }
+
+    function vBalanceOfAtTime(uint256 _tokenId, uint256 _t) external view override returns (uint256) {
+        require(_t <= block.timestamp, "Invalid timestamp");
```
eric6311, line 51: Does it need to forbid user querying the past voting power? Perhaps users just want to calculate the voting power by themselves?

charlesjhongc: Actually it avoid `_t` is in the future. Historical vBalance querying is available.

contracts/veLON.sol
charlesjhongc: Right, renamed as `blockRate` in 6bbc955

contracts/veLON.sol
eric6311: Oh sorry, I misunderstand the symbol haha

eric6311 (approved):

PR #71, opened 2022-08-03
Add veLON balance test

charlesjhongc:


contracts/test/veLON/Setup.t.sol
```diff
@@ -10,17 +10,17 @@ contract TestVeLON is Test {
...
+    address bob = vm.addr(uint256(2));
+    address alice = vm.addr(uint256(3));
+    address other = vm.addr(uint256(4));
 
     Lon lon = new Lon(address(this), address(this));
     veLON veLon;
 
-    uint256 constant DEFAULT_STAKE_AMOUNT = 1e18;
-    uint256 constant DEFAULT_LOCK_TIME = 14 days;
+    // set default stake amount to 10 years in order to have integer initial vBalance easily
```
eric6311, line 21: What does it mean setting default **"amount"** to 10 **"years"**?

charlesjhongc: To avoid division rounding. The first step of point calculation is get declining rate which equals `amount / maxDuration(1 year)`.

charlesjhongc: `10 years` is just a number and the real meaning depends on how it is interpreted. For example if it's interpreted as seconds, then it means 10 years in seconds.

eric6311 (approved):

zeqing-guo (approved):

PR #77, opened 2022-08-15
Remove PMM

charlesjhongc:
Remove PMM contracts and test written in typescript. Also, remove libs from 0x.

SDK upgrading is needed with new version of MMProxy. The interface is unchanged so as long as the 1271 wallet parse `bytes signature` using the same structure as used when signing, it should be working without any updates.


NIC619 (approved):

NIC619:
Finally!
Can we remove hardhat and typescript dependencies all together?

charlesjhongc:
I think hardhat can be removed as well. Will cleanup the rest in another PR.

pilagod (approved):
LGTM 😎

PR #78, opened 2022-08-16
Contracts refactoring

charlesjhongc:
- Add licenses
- Share constant lib
- Set function mutability
Clear compiler warning completely (test not included).

### DIFF
--- a/contracts/AllowanceTarget.sol
+++ b/contracts/AllowanceTarget.sol
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.7.6;
-
 import "@openzeppelin/contracts/utils/Address.sol";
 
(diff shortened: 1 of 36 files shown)


contracts/AllowanceTarget.sol
```diff
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.7.6;
-
```
pilagod, line 3: This blank line should be kept, right?

charlesjhongc: Yeah, I was trying to make it consistent across all contrasts. Thanks for the reminding. Fixed in 119500a

pilagod (approved):

NIC619 (approved):

PR #86, opened 2022-08-30
fix multicall result data

charlesjhongc:
If one of subcall is reverted during multicall process, the `result` offset will be changed in order to get the revert message. Therefore, copying the `success` and `result` into array should be done before changing the offset.

### DIFF
--- a/contracts/UserProxy.sol
+++ b/contracts/UserProxy.sol
@@ -145,14 +145,11 @@ contract UserProxy is Multicall {
     function toAMM(bytes calldata _payload) external payable {
         require(isAMMEnabled(), "UserProxy: AMM is disabled");
 
-        (bool callSucceed, ) = ammWrapperAddr().call{ value: msg.value }(_payload);
-        if (callSucceed == false) {
-            // Get the error message returned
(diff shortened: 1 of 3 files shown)


contracts/UserProxy.sol
```diff
@@ -145,14 +145,11 @@ contract UserProxy is Multicall {
     function toAMM(bytes calldata _payload) external payable {
         require(isAMMEnabled(), "UserProxy: AMM is disabled");
 
-        (bool callSucceed, ) = ammWrapperAddr().call{ value: msg.value }(_payload);
-        if (callSucceed == false) {
-            // Get the error message returned
+        (bool callSucceed, bytes memory result) = ammWrapperAddr().call{ value: msg.value }(_payload);
```
NIC619, line 148: Maybe we don't need to assign return data to `result`? Because we are still going to remove the prefixed byte (length of `result` data) in assembly (`revert(add(result, 0x20), returndatasize())`) and we are not using the `result` anyway.

NIC619 (approved):

pilagod (approved):

PR #88, opened 2022-09-05
Add strategy base contract

charlesjhongc:
Based on branch `add-l2-deposit-contract-new-based`.

The StrategyBase contract covers following basic aspects:
- Ownership management
- `spender` management
- Token allowance and WETH deposit (mainly for emergency purpose)
- onlyUserProxy() modifier
- `permStorage` address

### DIFF
--- a/contracts/interfaces/IStrategyBase.sol
+++ b/contracts/interfaces/IStrategyBase.sol
@@ -1,0 +1,17 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.7.0;
+
+interface IStrategyBase {
+    event UpgradeSpender(address newSpender);
+    event AllowTransfer(address spender);
+    event DisallowTransfer(address spender);
+    event DepositETH(uint256 ethBalance);
+
+    function upgradeSpender(address _newSpender) external;
+
+    function setAllowance(address[] calldata _tokenList, address _spender) external;
+
+    function closeAllowance(address[] calldata _tokenList, address _spender) external;
+
+    function depositETH() external;
+}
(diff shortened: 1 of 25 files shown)


contracts/interfaces/IStrategyBase.sol
```diff
@@ -0,0 +1,17 @@
...
+
+interface IStrategyBase {
+    event UpgradeSpender(address newSpender);
+    event AllowTransfer(address spender);
+    event DisallowTransfer(address spender);
+    event DepositETH(uint256 ethBalance);
+
+    function upgradeSpender(address _newSpender) external;
+
+    function setAllowance(address[] calldata _tokenList, address _spender) external;
```
changwu-tw, line 12: It looks more clean now. My only concern is that do all of our strategy contracts need the base contract? 

For example, if the strategy does not involve with financial products, does it need to have DepositETH? Or, since we will collect tokens in another contract in the future, do we still need Allow/DisallowTransfer? 

It might be ok for AMM/RFQ/LO contracts now, perhaps we can have different base contracts for specific usage later.

changwu-tw (approved):

contracts/interfaces/IStrategyBase.sol
charlesjhongc: Right, I thought about it but actually these function are not only just for raw ETH handling and fee collecting, they can support emergency needs as well (e.g. unexpected asset left in strategy contract). If a strategy contract doesn't have any payable function or token transfer, then it can assume that there will have very little chance for the needs of emergency pulling.

charlesjhongc:
Force pushed to discard stale commit.

pilagod (approved):

NIC619 (approved):

PR #95, opened 2022-09-16
strictly event emitted

ChiHaoLu:
- After overviewing the whole tokenlon contracts testing, I turn the all event unit testing to emitting strictly. Namely, the event testing must be `expectEmit(true,  true, true, true)`.
- Related Internal documentation: **Strictly Event Emitting Unit Testing in Tokenlon**

### DIFF
--- a/contracts/test/PermanentStorage.t.sol
+++ b/contracts/test/PermanentStorage.t.sol
@@ -93,7 +93,7 @@ contract PermanentStorageTest is Test {
 
     function testUpgradeAMMWrapper() public {
         assertEq(permanentStorage.ammWrapperAddr(), address(0));
-        vm.expectEmit(false, false, false, true);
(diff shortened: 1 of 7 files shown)


NIC619 (approved):

NIC619:
My understanding was that for the event topics not used (i.e., no` indexed` keywords used in the event), `expectEmit` will raise error if we expect them to emit but seems like I was wrong. Though I guess there's not much point providing options in `expectEmit` then, just defaulting to checking all topcis and value.

NIC619:
Also maybe not exposing links to our internal docs here?

ChiHaoLu:
> Also maybe not exposing links to our internal docs here?

Oh, that's a big mistake...

NIC619:
No worries

changwu-tw (approved):

PR #96, opened 2022-10-18
Reuse unit test for production network

charlesjhongc:
Refactor unit tests for stategy contracts to support deployed environment.
- Read address from env variables when the contracts is deployed already
- Set and prank `owner` explicitly


NIC619:
Checking that the deployment works seems like a one time thing to me and I remember we dropped the deployment tests back then. Why are we integrating this into our normal (unit) tests?
Given that it's only used when we are deploying, the complexity added to the testing code base does not seem to be worth it in my opinion.

contracts/test/RFQ.t.sol
```diff
@@ -41,25 +41,50 @@ contract RFQTest is StrategySharedSetup {
...
+            dai = IERC20(vm.envAddress("DAI_ADDRESS"));
+
+            allowanceTarget = AllowanceTarget(vm.envAddress("AllowanceTarget_ADDRESS"));
+            spender = Spender(vm.envAddress("Spender_ADDRESS"));
+            userProxy = UserProxy(payable(vm.envAddress("UserProxy_ADDRESS")));
+            permanentStorage = PermanentStorage(vm.envAddress("PermanentStorage_ADDRESS"));
+
+            rfq = RFQ(payable(vm.envAddress("RFQ_ADDRESS")));
+            owner = rfq.owner();
+            feeCollector = rfq.feeCollector();
```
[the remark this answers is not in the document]
pilagod: Same here. It's little inconvenient to add deployed version support for each tests.

charlesjhongc:
The purpose is to make sure new deployed strategy contracts have been set up correctly. It foucuses on whether parts of the upgrading process are missing or not have been done correctly. For example, forget to grant spender permission or the address in permanent storage is not updated. In other words, we have to make sure txs can be processed and settled after deploying or upgrading contracts.
There're different ways to make sure new deployed contract works. However, instead of having a dedicated test or scripts for this which needs to be maintainned separately, I think re-use uint test is more efficient thought it covers more than the needs at this stage. The params that is used during deploying or the state of deployed contracts should be verifyied as well and it will be convered in another ticket and PRs.

contracts/test/RFQ.t.sol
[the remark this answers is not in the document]
NIC619: I think if these duplication can be avoided and the codes are hidden somewhere where other people writing unit tests do not have to worry about, it's worth the change.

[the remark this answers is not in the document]
pilagod: I think the `Tokens` extraction is a nice refactoring 👍 In my opinion it would be better to apply the same technique to combine `loadDeployedSystemContracts` and `setUpSystemContracts` to the lower level util contract.

For me `setUpSystemContracts` doesn't imply anything about deploying, it is only saying it will somehow setup system contracts for us to do the testing. So I think embedding `loadDeployedSystemContracts` into it is not a problem.

.github/workflows/node.js.yml
```diff
@@ -40,6 +40,7 @@ jobs:
           yarn run compile
       - name: Test
         env:
+          deployed: false
```
pilagod, line 43: nit: Could env variable stick to all capital character convention?

envrc
```diff
@@ -0,0 +1,19 @@
+export deployed=false
+
+export AllowanceTarget_ADDRESS="0x62b546741A5D59733E1D7ee2f12dF3b790Fc1bb8"
```
pilagod, line 3: nit: Same as the env comment above. Use unified naming rule would improve the readability and decrease the mental overhead

contracts/test/RFQ.t.sol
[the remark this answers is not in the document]
charlesjhongc: Yeah I thought about it and felt ok with both direction. But RFQ can't have a clean split if these two merged. Checkout 0fbdfd299881d91be419610500f68500e14205ef.

envrc
charlesjhongc: Sure, didn't notice at the time. Fixed in e8aa620 & 2b1f74d.

contracts/test/utils/Tokens.sol
```diff
@@ -0,0 +1,49 @@
...
+            ankreth = IERC20(vm.envAddress("ANKRETH_ADDRESS"));
+        } else {
+            // load ERC20s using constant address
+            weth = IERC20(WETH_ADDRESS);
+            usdt = IERC20(USDT_ADDRESS);
+            usdc = IERC20(USDC_ADDRESS);
+            dai = IERC20(DAI_ADDRESS);
+            wbtc = IERC20(WBTC_ADDRESS);
+            lon = IERC20(LON_ADDRESS);
+            ankreth = IERC20(ANKRETH_ADDRESS);
```
NIC619, line 36: These should be freshly deployed otherwise they are the same as `DEPLOYED` and RFQ don't have to deploy them in `setUp()`.

contracts/test/forkMainnet/AMMWrapper/Setup.t.sol
```diff
@@ -11,22 +11,20 @@ import { getEIP712Hash } from "contracts-test/utils/Sig.sol";
...
     uint256 userPrivateKey = uint256(1);
     uint256 otherPrivateKey = uint256(2);
 
     address user = vm.addr(userPrivateKey);
-    address feeCollector = address(0x133701);
-    address relayer = address(0x133702);
+    address owner = makeAddr("owner");
+    address feeCollector = makeAddr("feeCollector");
+    address relayer = makeAddr("relayer");
+    address psOperator;
```
NIC619, line 23: This can also be generated by `makeAddr`?
```suggestion
    address psOperator = makeAddr("psOperator");
```

contracts/test/forkMainnet/AMMWrapper/Setup.t.sol
```diff
@@ -49,23 +47,34 @@ contract TestAMMWrapper is StrategySharedSetup {
...
-        ammQuoter = new AMMQuoter(
-            UNISWAP_V2_ADDRESS,
-            UNISWAP_V3_ADDRESS,
-            UNISWAP_V3_QUOTER_ADDRESS,
-            SUSHISWAP_ADDRESS,
-            BALANCER_V2_ADDRESS,
-            IPermanentStorage(permanentStorage),
-            address(weth)
-        );
+        if (vm.envBool("DEPLOYED")) {
```
NIC619, line 52: Maybe these if/else code blocks in each strategy's `setUp` can be moved into a `_setUpDeployedStrategy` hook (just like the `_deployStrategyAndUpgrade` hook)?

contracts/test/utils/Tokens.sol
charlesjhongc: For `DEPLOYED` case, it may be any network from testnet to L2 and therefore is different from forked mainnet. We can generalize the forked testing and always provide addresses before running a forked test but may cost another setup step when doing forked mainnet test.

contracts/test/forkMainnet/AMMWrapper/Setup.t.sol
charlesjhongc: There are some setup or assignment which is related to strategy contract and cannot be combined together. For example not every strategy contract has the `feeCollector` assignment.

pilagod (approved):

contracts/test/utils/Tokens.sol
NIC619: Yeah I forgot these are also used for forked tests...

I will write a setup that differentiate between different networks and setup tokens/system contracts accordingly.

contracts/test/forkMainnet/AMMWrapper/Setup.t.sol
NIC619: I think they can be done as a hook like `_deployStrategyAndUpgrade`? `_deployStrategyAndUpgrade` is also strategy specific.

In `StrategySharedSetup.sol`
```solidity
function setUpSystemContracts() {
    if (vm.envBool("DEPLOYED")) {
        ...
        _setUpDeployedStrategy();
    } else {
        ...
    }
}

function _setUpDeployedStrategy() internal virtual returns (address) {}
```

In `RFQ.t.sol`
```solidity
function setUp() {
    ...
}

function _setUpDeployedStrategy internal override returns (address) {
    rfq = RFQ(payable(vm.envAddress("RFQ_ADDRESS")));
    owner = rfq.owner();
    feeCollector = rfq.feeCollector();
}
```

contracts/test/forkMainnet/AMMWrapper/Setup.t.sol
charlesjhongc: Adjusted in 6a94edd8d0af285b2e3c4c4aa284e5dcbe77c4b1

contracts/test/forkMainnet/AMMWrapper/Setup.t.sol
charlesjhongc: Sounds good, adjusted in 551bf2fed94383cca8956ce924bb5c0bfa487b54.

But for the RFQ case, still have to overwrite tokens before `setUpSystemContracts()` when running in non-deployed environment. Do you think it should still be wrapped in a `_setupDeployedStrategy()` hook in RFQ or leave it in the setup if/else block?

NIC619: I think we can leave the token overwrite in `setUp` for now.

```solidity
function setUp() {
    if (!vm.envBool("DEPLOYED")) {
        // overwrite tokens with locally deployed mocks
        weth = IERC20(address(new MockWETH("Wrapped ETH", "WETH", 18)));
        ...
    }
    setUpSystemContracts();
    ...
}

function _setupDeployedStrategy() {
    rfq = RFQ(payable(vm.envAddress("RFQ_ADDRESS")));
    ...
}
```

NIC619 (approved):

pilagod (approved):

PR #100, opened 2022-10-25
add default env value for tests

charlesjhongc:
Add default env value in `package.json` so no need to set env variable before running normal tests.


NIC619 (approved):

pilagod:
Shall we also remove the flag in GitHub CI workflow?

pilagod (approved):

oneleo (approved):

PR #104, opened 2022-10-28
Feature RFQ adds  spendByRFQ option when fill

doublespending:


contracts/RFQ.sol
```diff
@@ -5,7 +5,8 @@ pragma abicoder v2;
 import "@openzeppelin/contracts/math/SafeMath.sol";
 import "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
 import "@openzeppelin/contracts/utils/Address.sol";
-
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
```
charlesjhongc: Keep the empty line to stay consistent.

contracts/RFQ.sol
```diff
@@ -144,19 +145,59 @@ contract RFQ is IRFQ, StrategyBase, ReentrancyGuard, SignatureValidator, BaseLib
...
+        address _tokenAddr,
+        address _recipient,
+        uint256 _amount,
+        bool _useSpenderForMaker
+    ) internal {
+        if (_useSpenderForMaker) {
+            spender.spendFromUserTo(_user, _tokenAddr, _recipient, _amount);
+            return;
+        }
+        uint256 balanceBefore = IERC20(_tokenAddr).balanceOf(_recipient);
```
charlesjhongc: We may need to review the needs of this before/after check under different situations.

contracts/test/RFQ.t.sol
```diff
@@ -513,6 +570,18 @@ contract RFQTest is StrategySharedSetup {
         bytes memory makerSig,
         bytes memory userSig
     ) internal view returns (bytes memory payload) {
-        return abi.encodeWithSelector(rfq.fill.selector, order, makerSig, userSig);
+        bytes4 selector = rfq.fill.selector;
```
charlesjhongc, line 573: Suggest to have another params for `_genFillPayload()` or can split into another payload getting function to be more specific instead of reading from the global `isMakerOnlyApproveRFQ` flag.

contracts/test/RFQ.t.sol
```diff
@@ -311,6 +325,13 @@ contract RFQTest is StrategySharedSetup {
         makerMakerAsset.assertChange(-int256(order.makerAssetAmount));
     }
 
+    function testFillDAIToUSDT_EOAUserAndEOAMaker_MakerOnlyApproveRFQ() public {
+        RFQLibEIP712.Order memory order = DEFAULT_ORDER;
+        // maker only approve RFQ
+        _makerOnlyApproveRFQ(order.makerAddr, order.makerAssetAddr);
+        testFillDAIToUSDT_EOAUserAndEOAMaker();
```
charlesjhongc, line 332: Not sure calling another test case is a good idea. I think it can be independent since the rest can be done in few lines.

charlesjhongc: Also, I think we can focus on allowance/settlement here so no need to test each combination of EOA/Contract signing case.

contracts/utils/LocalTransferERC20.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.7.6;
+pragma abicoder v2;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+
+// Modified from https://github.com/Uniswap/uniswap-v3-periphery/blob/v1.1.1/contracts/base/Multicall.sol
```
charlesjhongc, line 8: Remove?

contracts/RFQ.sol
NIC619: If we don't check balance match, then transfers of inflationary tokens would succeed. Since we now transfer all fee away from strategy contract, allowing inflationary tokens would not hurt us but the numbers would probably not match between expected amount we displayed to user and user's actual received amount.

charlesjhongc: Since maker only provide quotes to certain pair, inflationary tokens can be avoided in RFQ. I think there's no reason for a maker or user to skip our API but it's an assumption.

doublespending: if we abstract this to a common contract, check balance may be better?

contracts/utils/LocalTransferERC20.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.7.6;
+pragma abicoder v2;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+
+abstract contract LocalTransferERC20 {
```
pilagod, line 8: I feel this extraction is not worth it, unless we could also combine the use spender check to it. (in other words, the whole `_spendFromUserTo` body on RFQ)

But I have no good idea for the merge now 🥲 If other buddies have no ideas too, then I would suggest inlining `_localSpendFromUserTo` body in `_spendFromUserTo` on RFQ.

contracts/RFQ.sol
```diff
@@ -83,22 +84,18 @@ contract RFQ is IRFQ, StrategyBase, ReentrancyGuard, SignatureValidator, BaseLib
...
+        GroupedVars memory vars = _preHandleFill(_order, _mmSignature, _userSignature);
+        return _settle({ _order: _order, _vars: vars, _useSpenderForMaker: true });
+    }
 
-        return _settle(_order, vars);
+    function fillWithSpendOption(
+        RFQLibEIP712.Order calldata _order,
+        bytes calldata _mmSignature,
+        bytes calldata _userSignature,
+        SpendOption calldata _spendOption
```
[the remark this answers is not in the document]
NIC619: I'm fine with both.

contracts/utils/LocalTransferERC20.sol
NIC619: It should probably at least incorporate the balance check.

For consumeGasERC20 check, since USDT/LINK/BNT are still very popular old ERC20 tokens but the list have not been updated ever, I think we can maybe update the list in constructor (hardcode the list or not) once and for all?

```solidity
constant consumeGasERC20 = [USDT_ADDR, LINK_ADDR, BNT_ADDR]

contract LocalTransferERC20 {
    constructor () {
        for addr in consumeGasERC20 {
            consumeGasERC20Tokens[addr] = true;
        }
    }
    function _localSpendFromUserTo(...) internal {
        ...
        uint256 gasStipend = consumeGasERC20Tokens[_tokenAddr] ? 80000 : gasleft();
        ...
    }
}
```

doublespending: what about token black list?

NIC619: The blacklist is never used.😂 I think we can remove it.

contracts/RFQ.sol
charlesjhongc: Then the argument is based on that RFQ must share a common spending function from other contracts. But if we replace this inheritance by one line ERC20 transfering, we can save some gas in RFQ.

contracts/RFQ.sol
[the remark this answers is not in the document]
charlesjhongc: Vote for just boolean.

charlesjhongc (approved):
LGTM

NIC619 (approved):

pilagod (approved):

PR #106, opened 2022-10-31
Add doc for Tokenlon

charlesjhongc:


doc/README.md
```diff
@@ -0,0 +1,43 @@
...
+## MarketMakerProxy
+
+# Trading Strategy Contracts
+
+-   AMMWrapper & AMMQuoter
+-   [RFQ (MarketMakerProxy)](./strategies/RFQ.md)
+-   Limit Order
+-   L2Deposit
+
+# LON Token peripherals
```
pilagod, line 36: nit: Capitalize `peripherals` 😁

pilagod (commented):
The structure LGTM 👍

doc/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+3. [LON Token Peripherals](#LON-Token-Peripherals)
+
+![image info](../tokenlon_architecture.png)
+
+# Protocol Infrastructure
+
+## UserProxy & Tokenlon
+
+User proxy is the entry of the whole procotol while Tokenlon is a transparent upgradeable proxy of it. User proxy navigate users to a specific strategy contract. Meanwhile, it has a multicall entry which allows batching calls between differnt strategy contracts in a single transaction.
```
doublespending, line 19: Strategy contracts are not mentioned later in the article? And
shall we explain the concept of strategy contract?

doc/README.md
```diff
@@ -0,0 +1,51 @@
...
+## MarketMakerProxy
+
+MarketMakerProxy is an example implementation of a EIP-1271 compatable contract wallet. It helps market makers or advanced users to onboard Tokenlon more easily.
+
+# Executing Contracts
+
+-   AMMWrapper & AMMQuoter
+-   [RFQ](./executing/RFQ.md)
+-   Limit Order
+-   L2Deposit
```
doublespending, line 42: Shall we take L2Deposit apart from Executing Contract, list it under `L2 Related Contracts` or `Miscellaneous Contracts` and finally rename `Executing Contracts` to `Trading Contracts` or `Strategy Contracts`?

doc/README.md
changwu-tw: how about having the structure like this? 

```
/docs 
├── README.md
├── advanced
│   └── universal_permit.md
├── architecture
│   └── protocol.md
├── peripheral
│   └── LON.md
└── strategy
    └── RFQ.md
```

doc/README.md
charlesjhongc: Yeah I think how strategy contracts are integrated should be at least mentioned here. Will fix.

doc/README.md
charlesjhongc: Then we still need a way to describe those contracts as a whole. `Miscellaneous Contracts` and `Trading Contracts` can be subsets of that.

NIC619:
I think one core proposition of Tokenlon is to served as a convenient and safe entry for user to access different protocols on-chain.
- Convenient because users only need to approve Tokenlon once (one for each token)
- Safe because Tokenlon can guard user against rogue contracts (user don't have to worry about his approvals for rouge contracts)

User can use different "strategy" to do different things conveniently and safely. Strategy can be simple or complex: it can be as simple as a dust sweeper, an airdrop tool, an AMM swap, a bridge to L2 to a complex Defi protocol combo.
To me "strategy" is not tied to "trading". I would prefer "strategy" over "executing" because "executing" seem too opaque to me personally.

doc/README.md
charlesjhongc: Currently protocol doc is in `README.md` cause it can be displayed at the beginning and parts for permit will be added in this place once feature is merged. `advanced` is a not clear for me what should be inside if I'm not familiar with codebas.  `Peripherals` will have a sub cat just like `strategy`.

charlesjhongc:
Adjusting `strategy` naming in f9a5173

changwu-tw (approved):

pilagod (approved):

PR #113, opened 2022-11-07
refac: rfq foundry test with named parameter

oneleo:
# RFQ Foundry test with Named Parameter

1. Only change the RFQ Foundry test contract to named parameter style.
2. Only change the calling 'internal function' to named parameter style.
3. Verify command (Not necessary):
```shell
tokenlon-contracts % DEPLOYED=false forge test -vvv --match-path 'contracts/test/RFQ.t.sol'
```

### DIFF
--- a/contracts/test/RFQ.t.sol
+++ b/contracts/test/RFQ.t.sol
@@ -443,7 +443,7 @@ contract RFQTest is StrategySharedSetup {
         bytes memory userSig = _signFill({ privateKey: userPrivateKey, order: order, sigType: SignatureValidator.SignatureType.EIP712 });
         bytes memory payload = _genFillPayload({ order: order, makerSig: makerSig, userSig: userSig });
 
-        _expectEvent({ order: order });
+        _expectEvent(order);
         vm.prank(user, user); // Only EOA
         userProxy.toRFQ(payload);
     }


contracts/test/RFQ.t.sol
```diff
@@ -293,65 +293,65 @@ contract RFQTest is StrategySharedSetup {
...
         userProxy.toRFQ(payload);
 
-        userTakerAsset.assertChange(-int256(order.takerAssetAmount));
-        receiverMakerAsset.assertChange(int256(order.makerAssetAmount));
-        makerTakerAsset.assertChange(int256(order.takerAssetAmount));
-        makerMakerAsset.assertChange(-int256(order.makerAssetAmount));
+        userTakerAsset.assertChange({ expectedChange: -int256(order.takerAssetAmount) });
+        receiverMakerAsset.assertChange({ expectedChange: int256(order.makerAssetAmount) });
+        makerTakerAsset.assertChange({ expectedChange: int256(order.takerAssetAmount) });
+        makerMakerAsset.assertChange({ expectedChange: -int256(order.makerAssetAmount) });
```
pilagod, line 311: nit: I think the function name is already presenting enough meaning for reader, maybe only one argument case could get rid of named parameter?

oneleo: Sure, I think it can be got rid when understand the usage of Balance Snapshot.Snapshot.
What do others think?

NIC619: Yeah, we only need it for functions with multiple parameters.

NIC619 (approved):

contracts/test/RFQ.t.sol
```diff
@@ -439,11 +439,11 @@ contract RFQTest is StrategySharedSetup {
...
         RFQLibEIP712.Order memory order = DEFAULT_ORDER;
-        bytes memory makerSig = _signOrder(makerPrivateKey, order, SignatureValidator.SignatureType.EIP712);
-        bytes memory userSig = _signFill(userPrivateKey, order, SignatureValidator.SignatureType.EIP712);
-        bytes memory payload = _genFillPayload(order, makerSig, userSig);
+        bytes memory makerSig = _signOrder({ privateKey: makerPrivateKey, order: order, sigType: SignatureValidator.SignatureType.EIP712 });
+        bytes memory userSig = _signFill({ privateKey: userPrivateKey, order: order, sigType: SignatureValidator.SignatureType.EIP712 });
+        bytes memory payload = _genFillPayload({ order: order, makerSig: makerSig, userSig: userSig });
 
-        _expectEvent(order);
+        _expectEvent({ order: order });
```
pilagod, line 446: nit: This is also a single parameter function 😆

pilagod (approved):

charlesjhongc (approved):

contracts/test/RFQ.t.sol
oneleo: OK ~

NIC619 (approved):

pilagod (approved):

charlesjhongc (approved):

PR #114, opened 2022-11-09
refac: move the signSpendWithPermit function in foundry test

oneleo:
1. Centralize the signSpendWithPermit function into Permit.sol to reduce the amount of code.
2. Currently only the RFQ.t.sol and Spender.t.sol files are modified.
3. Verify command:
```shell
tokenlon-contracts % DEPLOYED=false forge test -vvv --match-path 'contracts/test/Spender.t.sol'
tokenlon-contracts % DEPLOYED=false forge test -vvv --match-path 'contracts/test/RFQ.t.sol'
```

### DIFF
--- a/contracts/test/utils/Permit.sol
+++ b/contracts/test/utils/Permit.sol
@@ -1,0 +1,41 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.7.6;
+pragma abicoder v2;
+
+import "@openzeppelin/contracts/cryptography/ECDSA.sol";
+import "forge-std/Test.sol";
+import "contracts/Spender.sol";
+import "contracts/utils/SpenderLibEIP712.sol";
+import "contracts/utils/SignatureValidator.sol";
+import { getEIP712Hash } from "contracts-test/utils/Sig.sol";
+
+contract Permit is Test {
+    function signSpendWithPermit(
+        uint256 privateKey,
+        SpenderLibEIP712.SpendWithPermit memory spendWithPermit,
+        Spender spenderContract,
+        SignatureValidator.SignatureType sigType
+    ) internal returns (bytes memory sig) {
+        uint256 SPEND_WITH_PERMIT_TYPEHASH = 0x52718c957261b99fd72e63478d85d1267cdc812e8249f5a2623566c1818e1ed0;
+        bytes32 structHash = keccak256(
+            abi.encode(
+                SPEND_WITH_PERMIT_TYPEHASH,
+                spendWithPermit.tokenAddr,
+                spendWithPermit.requester,
+                spendWithPermit.user,
+                spendWithPermit.recipient,
+                spendWithPermit.amount,
+                spendWithPermit.actionHash,
+                spendWithPermit.expiry
+            )
+        );
+        bytes32 spendWithPermitHash = getEIP712Hash(spenderContract.EIP712_DOMAIN_SEPARATOR(), structHash); //_getEIP712Hash(structHash);
+        if (sigType == SignatureValidator.SignatureType.Wallet) {
+            (uint8 v, bytes32 r, bytes32 s) = vm.sign(privateKey, ECDSA.toEthSignedMessageHash(spendWithPermitHash));
+            sig = abi.encodePacked(r, s, v, uint8(sigType)); // new signature format
+        } else {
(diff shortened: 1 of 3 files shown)


contracts/test/utils/Permit.sol
```diff
@@ -0,0 +1,41 @@
...
+                spendWithPermit.tokenAddr,
+                spendWithPermit.requester,
+                spendWithPermit.user,
+                spendWithPermit.recipient,
+                spendWithPermit.amount,
+                spendWithPermit.actionHash,
+                spendWithPermit.expiry
+            )
+        );
+        bytes32 spendWithPermitHash = getEIP712Hash(spenderContract.EIP712_DOMAIN_SEPARATOR(), structHash); //_getEIP712Hash(structHash);
```
charlesjhongc, line 32: What does the comment mean?

contracts/test/utils/Permit.sol
```diff
@@ -0,0 +1,41 @@
...
+import "contracts/Spender.sol";
+import "contracts/utils/SpenderLibEIP712.sol";
+import "contracts/utils/SignatureValidator.sol";
+import { getEIP712Hash } from "contracts-test/utils/Sig.sol";
+
+contract Permit is Test {
+    function signSpendWithPermit(
+        uint256 privateKey,
+        SpenderLibEIP712.SpendWithPermit memory spendWithPermit,
+        Spender spenderContract,
```
charlesjhongc, line 16: Pass EIP-712 domain separator should be enough.

contracts/test/utils/Permit.sol
```diff
@@ -0,0 +1,43 @@
...
+                spendWithPermit.requester,
+                spendWithPermit.user,
+                spendWithPermit.recipient,
+                spendWithPermit.amount,
+                spendWithPermit.actionHash,
+                spendWithPermit.expiry
+            )
+        );
+        bytes32 spendWithPermitHash = getEIP712Hash(domainSeparator, structHash);
+        if (sigType == SignatureValidator.SignatureType.EIP712 || sigType == SignatureValidator.SignatureType.WalletBytes32) {
```
charlesjhongc, line 33: What's the reason to include `WalletBytes32` here?

oneleo: Some Foundry Test need it, like this:
[https://github.com/consenlabs/tokenlon-contracts/blob/master/contracts/test/RFQ.t.sol#L275](https://github.com/consenlabs/tokenlon-contracts/blob/master/contracts/test/RFQ.t.sol#L275)

contracts/test/utils/Permit.sol
```diff
@@ -0,0 +1,43 @@
...
+pragma abicoder v2;
+
+import "@openzeppelin/contracts/cryptography/ECDSA.sol";
+import "forge-std/Test.sol";
+import "contracts/Spender.sol";
+import "contracts/utils/SpenderLibEIP712.sol";
+import "contracts/utils/SignatureValidator.sol";
+import { getEIP712Hash } from "contracts-test/utils/Sig.sol";
+
+contract Permit is Test {
```
pilagod, line 12: From my understanding this permit util collection doesn't depend on the Test base contract, and also has no setup needs. 

Maybe it is enough to construct it as plain library to reduce inheritance.

oneleo: Originally wanted to put signSpendWithPermit() function in the sig.sol file,
but the vm.sign() function should still inherit from test.sol

pilagod: My bad 🤦‍♂️

pilagod:
Should we apply the named parameter rule for the new code added, would like to hear everyone's opinion. 

@charlesjhongc @oneleo

oneleo:
> Should we apply the named parameter rule for the new code added? I would like to hear everyone's opinion.
> 
> @charlesjhongc @oneleo

Originally, I think the spender_permit_feature branch can simply do the spendWithPrpermit function, and the named parameter can be changed one by one in the master branch.

pilagod:
No problem, @charlesjhongc has also synced with me more about the context and schedule, that's go without named parameter first 💪

pilagod (approved):

charlesjhongc (approved):

PR #118, opened 2022-11-15
Adjust RFQ weth handling

charlesjhongc:
Allow WETH as taker asset. Wrap raw ETH only if taker specified and `msg.value` match the taker amount.

### DIFF
--- a/contracts/utils/RFQLibEIP712.sol
+++ b/contracts/utils/RFQLibEIP712.sol
@@ -56,12 +56,12 @@ library RFQLibEIP712 {
         );
     }
 
-    bytes32 public constant FILL_WITH_PERMIT_TYPEHASH = 0x4ea663383968865a4516f51bec2c29addd1e7cecce5583296a44cc8d568cad09;
+    bytes32 public constant FILL_TYPEHASH = 0x9b822fe3212f0c2a34b120a750281383593d621b7cee5888ef0bdc840bf6ff2c;
 
     /*
         keccak256(
             abi.encodePacked(
-                "fillWithPermit(",
+                "fill(",
                 "address makerAddr,",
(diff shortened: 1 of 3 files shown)


contracts/utils/RFQLibEIP712.sol
```diff
@@ -56,12 +56,12 @@ library RFQLibEIP712 {
         );
     }
 
-    bytes32 public constant FILL_WITH_PERMIT_TYPEHASH = 0x4ea663383968865a4516f51bec2c29addd1e7cecce5583296a44cc8d568cad09;
+    bytes32 public constant FILL_TYPEHASH = 0x9b822fe3212f0c2a34b120a750281383593d621b7cee5888ef0bdc840bf6ff2c;
```
pilagod, line 59: This refactor seems to affect current clients. Should we leave it as before until we decide to do a breaking upgrade?

NIC619:
I think one problem could be user thinks ETH and WETH are different as it is treated as two different taker asset but when it comes to maker asset WETH will just be converted to ETH, e.g., user trades WETH to USDT just fine but when he trades USDT to WETH it will be converted to ETH.

charlesjhongc:
Oh I forgot to adjust the maker asset part. My original thought was that WETH and ETH is different and RFQ does wrap/unwrap only if taker or maker specify it. @NIC619 What do you think if similar logic is added on the maker asset side?

NIC619:
I think this would be a breaking change for our backend and maker? As they have to differentiate between ETH and WETH when signing orders. Maybe we should check with backend team and maker to make sure it won't become a obstacle to our contract upgrade schedule?

charlesjhongc:
Will be covered by RFQv2 contract.

PR #119, opened 2022-11-16
feat: remove spendFromUser() func in Spender.sol

oneleo:
1. The spendFromUser() function in the Spender.sol contract was replaced by the new spendFromUserToWithPermit() function, so the spendFromUser() was removed and the associated Foundry Test was also removed.

2. Verify command:
```shell
DEPLOYED=false forge test -vvv --match-path 'contracts/test/Spender.t.sol'
```


charlesjhongc:
Let's merge to feature branch first.

PR #120, opened 2022-11-20
Limit order with old spender

charlesjhongc:
- Checkout from `v5.3.0`
- Fix Multicall result assignment
- Remove LimitOrder fee cap
- Adopt old version spender


changwu-tw:
I think this is the v5.2.0 with the limit order, but using old spender contract.

charlesjhongc:
Right, but `v5.2.0` was just a snapshot and was too old so I checked out from `v5.3.0`.

pilagod (approved):

PR #124, opened 2022-11-23
Add limit order doc

charlesjhongc:


doc/strategies/LimitOrder.md
```diff
@@ -0,0 +1,28 @@
+# LimitOrder
+
+The `LimitOrder` contract provides order book style trading functionalities which is similar to what centralized exchange does. Orders can be queried from the off-chain order book system and then be filled on chain. Also, it supports order cancelling and partially fill. The `LimitOrder` contract shares the `AllowanceTarget` design from Tokenlon which allow users to create an order without depositing tokens first. Tokens are transferred between taker and maker only when the order gets filled on chain. This design provides the certainty of price for maker while maintaining filling flexibilities for taker at the same time.
```
NIC619, line 3: I don't think the idea for `AllowanceTarget` is `allow users to create an order without depositing tokens first`?  Maybe you are referring to some other contract?

doc/strategies/LimitOrder.md
```diff
@@ -0,0 +1,28 @@
...
+
+Besides the traditional taker scenario, an order can also be satisfied by liquidity from a supported AMM protocol. It relies on someone(EOA) trying to match the price of AMM protocols and the opened orders. In order to incentivize relayers to participate, the difference of the price is considered as the profit of a relayer.
```
NIC619, line 5:
```suggestion
Besides the traditional taker scenario, an order can also be satisfied by liquidity from a supported AMM protocol. It relies on relayer(EOA) to match the price of AMM protocols and the opened orders. In order to incentivize relayers to participate, the difference of the price is considered as the profit of a relayer.
```

doc/strategies/LimitOrder.md
```diff
@@ -0,0 +1,28 @@
...
+
+To avoid multiple filling transactions for the same order colliding because insufficient amount is left, the coordinator design is introduced. The idea is that every fill should be signed by a coordinator so the available amount can be calculated as early as possible.
```
NIC619, line 7:
```suggestion
To avoid multiple filling transactions for the same order colliding, the coordinator design is introduced. The idea is that every fill should be signed by a coordinator so all the fills can be executed successfully.
```

doc/strategies/LimitOrder.md
```diff
@@ -0,0 +1,28 @@
...
+| makerTokenAmount | uint256 | The amount of the maker token.                                                  |
+| takerTokenAmount | uint256 | The amount of the taker token.                                                  |
+| maker            | address | The address of the maker.                                                       |
+| taker            | address | The address of the taker. Set to zero address if no specific taker is assigned. |
+| salt             | uint256 | A random number included in an order to avoid replay attack.                    |
+| expiry           | uint64  | The timestamp of the expiry.                                                    |
+
+## Order Cancelling
+
+If a maker wants to cancel an opened order, a special signature is required. More specifically, the maker should modify the original order with `takerTokenAmount` set to zero and sign it. By submitting cancelling signature to the contract, the original order is marked as canceled which can not by filled anymore.
```
NIC619, line 24:
```suggestion
If a maker wants to cancel an opened order, a special signature is required. More specifically, the maker should modify the original order with `takerTokenAmount` set to zero and sign it. By submitting cancelling signature to the contract, the original order is marked as canceled which can not be filled anymore.
```

doc/strategies/LimitOrder.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Fee
+
+There are totally three different fee factors in `LimitOrder` contract. For `fillByTrader` scenario, `makerFeeFactor` and `takerFeeFactor` can be applied to calculate two kinds of fees. For `fillByProtocol` case, the `takerFeeFactor` is applied to the taker asset as well. However, there's another `profitFeeFactor` for the profit of the relayer. All of the fees will be transferred to `feeCollector` during the settlement.
```
NIC619, line 28:
```suggestion
There are totally three different fee factors in `LimitOrder` contract. For `fillByTrader` scenario, `makerFeeFactor` and `takerFeeFactor` will be applied to calculate two kinds of fees. For `fillByProtocol` case, only the `takerFeeFactor` is applied to the taker asset. However, there's another `profitFeeFactor` to calculate the profit for the relayer. All of the fees will be transferred to `feeCollector` during the settlement.
```

doc/strategies/LimitOrder.md
charlesjhongc: Users have to deposit tokens first before creating an order in CEX but not in Tokenlon.

doc/strategies/LimitOrder.md
charlesjhongc: `Relayer` may cause reader thinks that there're only a group of accounts are able to do that. So I used `someone` to emphasize everyone can participate.

doc/strategies/LimitOrder.md
charlesjhongc: I think the reason regarding to the left amount should be mentioned here.

doc/strategies/LimitOrder.md
charlesjhongc: I think this can be left for formal grammar correction.

doc/strategies/LimitOrder.md
NIC619: Maybe we can just refer the design to DEX or Tokenlon? so reader doesn't have to look into `AllowanceTarget` contract to know what the "design" is as we didn't mention this "design" in the doc before.

doc/strategies/LimitOrder.md
NIC619: `relayer` is mentioned in the next paragraph: `... In order to incentivize relayers to participate, the difference of the price is considered as the profit of a relayer.`
If it is not mentioned here then readers won't know to what does the `relayer` refer.

doc/strategies/LimitOrder.md
NIC619: Sure. How about `To avoid multiple filling transactions for the same order at the same time and some of them failed due to insufficient amount, the coordinator design is introduced.`
Because `insufficient amount is left` is not the reason why `transactions colliding`. It's `transactions colliding and so some of them fail due to insufficient amount`.

pilagod (approved):

doc/strategies/LimitOrder.md
charlesjhongc: Actually that's my major purpose. Let me rephrase in a way.

charlesjhongc: Adjusted in ca3f142

doc/strategies/LimitOrder.md
charlesjhongc: Adjusted in ca3f142

pilagod (approved):

NIC619 (approved):

PR #132, opened 2022-12-21
Update configs and libraries

NIC619:


.github/workflows/tokenlon-contracts.yml
```diff
@@ -20,6 +17,11 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
+      - name: "Cache RPC Responses"
+        uses: "actions/cache@v3"
+        with:
+          path: "~/.foundry/cache/rpc/mainnet/15451000"
+          key: "${{ runner.os }}-mainnet-15451000"
```
pilagod, line 24: Q: Is this for CI performance improvement? 

I think it would be better to cache whole directory because it might be easy to miss updating the block number here when we change our target block in test. (But I not sure if it is possible to cache the whole directory. 😂)

test/LONStaking/RageExit.t.sol
```diff
@@ -100,14 +98,14 @@ contract TestLONStakingRageExit is TestLONStaking {
...
         for (uint256 i = 0; i < stakeAmounts.length; i++) {
-            uint256 stakeAmount = stakeAmounts[i];
-            vm.assume(stakeAmount > 0);
-            totalLONAmount = totalLONAmount.add(stakeAmount);
-            vm.assume(totalLONAmount <= lon.cap());
+            uint256 itemsLeft = stakeAmounts.length - i - 1;
+            // If stake amount is set to `lon.cap().sub(totalLONAmount)`, rest of the stake amount will all be zero and become invalid.
+            // So an additional `itemsLeft` is subtracted from the max value of the current stake amount
+            stakeAmounts[i] = bound(stakeAmounts[i], 1, lon.cap().sub(totalLONAmount).sub(itemsLeft));
+            totalLONAmount = totalLONAmount.add(stakeAmounts[i]);
```
pilagod, line 108: I am still not getting the reason to subtract additional number, especially it is an array index but subtraction from amount, which looks weird to me. 🤔 Could you provide a detailed example for it?

pilagod: Another thing is that these comments and actions seem to spread across multiple files which might incur maintaining problem in the future. It would be better if it could maybe be collected into an util to share among cases.

.github/workflows/tokenlon-contracts.yml
NIC619: Good point. Let me try.

test/LONStaking/RageExit.t.sol
NIC619: We need the stake amount to greater than `1` (and the sum of all the stake amount to be less than `lon.cap()`). So when bounding stake amount in each iteration, we need to make sure that rest of the stake amounts can still be greater than `1` and that's why the max stake amount is `lon.cap().sub(totalLONAmount).sub(itemsLeft)`. `.sub(itemsLeft)` is there to make sure rest of the stake amount can be greater than `1` (if those stake amounts are all `1`, their sum is `itemsLeft`).

NIC619: Maybe I should rename `itemsLeft` to `numStakeAmountLeft` and rewrite stake amount bound as `stakeAmounts[i] = bound(stakeAmounts[i], MIN_STAKE_AMOUNT, lon.cap().sub(totalLONAmount).sub(numStakeAmountLeft  * MIN_STAKE_AMOUNT));`

charlesjhongc:
The scripts in the private repo require typescript artifacts. Is there any way that the private hardhat one can integrate with the foundry-only code base?

NIC619:
@charlesjhongc Adjusted configs in private repo, compiling contracts and running script worked: https://github.com/consenlabs/tokenlon-contracts-private/pull/396

test/LONStaking/RageExit.t.sol
pilagod: Thanks for the explanation, it's much clearer 👍 I think the new suggestion you gave is more meaningful to me. We can go for it.

foundry.toml
```diff
@@ -4,22 +4,17 @@ test = 'test'                                                 # the test directo
...
 optimizer_runs = 200                                          # the number of optimizer runs
-
+ignored_error_codes = []                                      # a list of ignored solc error codes
 block_number = 10000
 block_timestamp = 1640966400                                  # 2022/01/01 00:00
 initial_balance = '0xffffffffffffffffffffffff'                # the initial balance of the test contract
 gas_limit = 9223372036854775807                               # the gas limit in tests
 gas_price = 0                                                 # the gas price (in wei) in tests
 
-verbosity = 3
```
charlesjhongc: Should not be removed?

charlesjhongc (dismissed):

foundry.toml
NIC619: Fixed in 1142793

test/LONStaking/RageExit.t.sol
NIC619: Fixed in 1142793

pilagod (approved):
I feel that fuzzing tests are complicate to write and incur much maintenance cost. In the future I think we could try our best to limit the fuzzing test count, to use them only on few core and must-not-go-wrong scenario.

test/LONStaking/RageExit.t.sol
NIC619: I moved the repeated bound function into util functions in da94bdf

NIC619:
Agree that fuzzing tests are not easy to maintain. And for most of our contracts they simply act as an interface/bridge to swap user's asset or bridge it to L2 so we don't need to worry about user asset getting stolen from our contracts. But for contracts like LONStaking or veLON where huge amount of money is locked in the contracts, I think it's very crucial to validate the contracts functionality with as many tools/tests as possible.

pilagod (dismissed):

charlesjhongc (dismissed):

NIC619:
@pilagod @charlesjhongc My bad. I resolved the conflicts by rebase and force push.

pilagod (approved):
gogogo

charlesjhongc (approved):

PR #135, opened 2023-01-10
Check EIP712 sigs in tests

NIC619:
1. Run SDK signing script (`sdkSign.ts`) first which writes signature generated to json
2. Run Foundry signing script (`foundrySign.s.sol`) which compares signature generated against the one generated in step 1.


pilagod:
Would you consider re-implementing get data hash logic in tests? (in other words, don't use library to get order hash in tests)

Imo additional signature check by other language it's not easy to maintain, with linear-growth effort for any new sig types.

NIC619:
> Would you consider re-implementing get data hash logic in tests? (in other words, don't use library to get order hash in tests)

No problem, I will move them to tests.

charlesjhongc:
Not sure the reason to verify SDK signing result in the contract repo. I think it should be considered as a part of integration test which focusing on the whole payload.

NIC619:
This is verify that EIP712 payload is correctly formed in the contracts.

test/forkMainnet/AMMWrapper/Signing.t.sol
```diff
@@ -0,0 +1,54 @@
...
+        address ammWrapperAddr = abi.decode(vm.parseJson(ammWrapperPayloadJson, "AMMWrapper"), (address));
+        uint256 userPrivateKey = abi.decode(vm.parseJson(ammWrapperPayloadJson, "signingKey"), (uint256));
+        bytes memory sig = _signAMMTrade(ammWrapperAddr, userPrivateKey, order);
+
+        bytes memory expectedSig = abi.decode(vm.parseJson(ammWrapperPayloadJson, "expectedSig"), (bytes));
+        require(keccak256(sig) == keccak256(expectedSig), "Not expected AMM order sig");
+    }
+
+    function _signAMMTrade(
+        address ammWrapperAddr,
```
NIC619, line 45: Need `_signAMMTrade` because in test AMMWrapper is freshly deployed while we need fixed AMMWrapper address in signing payload for sdk to sign.

test/forkMainnet/AMMWrapper/Signing.t.sol
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.7.6;
+
+import "test/utils/StrategySharedSetup.sol";
+
+function computeEIP712DomainSeparator(address verifyingContract) returns (bytes32) {
+    uint256 CHAIN_ID = 1;
```
NIC619, line 7: Chain ID is also fixed so Foundry or sdk don't have to coordinate on which network to test against. It's just pure signing, network environment is irrelevant.

test/forkMainnet/AMMWrapper/Signing.t.sol
pilagod: Oh I didn't realize this situation 🥲

NIC619: Me neither until I was writing it😅.

pilagod (dismissed):
I have no better idea to organize it at this moment, so I just approve it first for now. 🙌

NIC619:
The bottom line is that we have no way to validate if a 712 signature payload format is correct using solidity/foundry alone. So it's either this, importing third party tools to validate it in this repo OR keep this repo "clean" and only find out the format is wrong when FE/BE teams are integrating the contracts.

NIC619:
I'm fine with either one since the problem will be caught before deploying on production so the impact is limited. I'm also open to any other ideas about how to validate it.

test/signing/sdkSign.ts
```diff
@@ -0,0 +1,188 @@
...
+            signer: signer,
+            verifyingContract: limitOrderPayloadJson.LimitOrder,
+        },
+    )
+    limitOrderPayloadJson["expectedAllowFillSig"] = limitOrderAllowFillSig
+
+    fs.writeFileSync(LimitOrderPath, JSON.stringify(limitOrderPayloadJson, null, 2))
+}
+
+async function signL2Deposit(signer: Wallet) {
```
NIC619, line 135: Note: we need a release from `tokenlon-contract-lib-js` which includes the L2Deposit signing helpers. cc @pilagod

NIC619:
@pilagod @charlesjhongc I've added payloads and tests for RFQ, LimitOrder and L2Deposit. If you are okay with them, I will move the ts signing scripts to private repo.

test/signing/sdkSign.ts
pilagod: @NIC619 just upgrade the library to v0.0.2 💪

https://www.npmjs.com/package/@tokenlon/contracts-lib

pilagod:
I have no other questions 🫡

charlesjhongc:
Signing payload looks good.

NIC619:
@pilagod @charlesjhongc I removed ts signing script and `@tokenlon/contracts-lib`. After this PR is merged, I will update submodule in private repo and add signing scripts there.

pilagod (approved):

charlesjhongc (approved):

PR #136, opened 2023-01-10
Try foundry fmt

NIC619:


contracts/interfaces/ILimitOrder.sol
```diff
@@ -32,12 +32,7 @@ interface ILimitOrder is IStrategyBase {
     /// @param recipient The address of the recipient which will receive tokens from maker
     /// @param fillReceipt Contains details of this single fill
     event LimitOrderFilledByTrader(
-        bytes32 indexed orderHash,
-        address indexed maker,
-        address indexed taker,
-        bytes32 allowFillHash,
-        address recipient,
-        FillReceipt fillReceipt
+        bytes32 indexed orderHash, address indexed maker, address indexed taker, bytes32 allowFillHash, address recipient, FillReceipt fillReceipt
```
NIC619, line 35: This creates different result from `prettier` and `fmt`: `prettier` has a problem but `fmt` does not, though this line does not exceed 160 characters...so I'm not sure why `prettier` is complaining.

NIC619: And this difference from `fmt` results in it updating many function and event declarations. This one is one of the many examples.

NIC619:
`prettier` will convert into multiline if number of function param exceed three, even if line length hasn't reach 160 limit.

pilagod:
Weird unwritten rules 😅 

Overall I think prettier is a better fit to our preference for now. We can delay the foundry fmt integration and keep tracking its updates until it is better enough 💪

PR #137, opened 2023-01-11
Read addresses from config in tests

NIC619:


pilagod:
What about putting mainnet addresses in file as well? Then every network would share same way to organize addresses. (Now it seems for mainnet it has to update the sol file, while for other networks it has to update json files)

test/forkMainnet/AMMQuoter.t.sol
```diff
@@ -12,14 +12,14 @@ contract AMMQuoterTest is StrategySharedSetup {
 
     AMMQuoter ammQuoter;
 
-    address DEFAULT_MAKER_ADDR = UNISWAP_V2_ADDRESS;
-    address DEFAULT_TAKER_ASSET_ADDR = DAI_ADDRESS;
-    address DEFAULT_MAKER_ASSET_ADDR = USDT_ADDRESS;
+    address DEFAULT_MAKER_ADDR;
+    address DEFAULT_TAKER_ASSET_ADDR;
+    address DEFAULT_MAKER_ASSET_ADDR;
```
NIC619, line 17: Can not assign default value to address variables here as the addresses do not have default values.

pilagod: Oh I didn't realize that addresses can only be used after constructor called 🥲 But if we keep previous version, it will always get the mainnet addresses no matter in which network, since the override happens after constructor called, right?

NIC619: Yes, it will always default to mainnet addresses.

> Oh I didn't realize that addresses can only be used after constructor called 🥲

Me neither until I was writing it...😂😅

pilagod:
Propose another possible solution to make address usage more convenient

https://github.com/consenlabs/tokenlon-contracts/compare/add_test_network_filter...add_test_network_filter_trial

test/utils/Addresses.sol
```diff
@@ -1,38 +1,105 @@
...
+        UNISWAP_V3_QUOTER_ADDRESS = abi.decode(vm.parseJson(file, "UNISWAP_V3_QUOTER_ADDRESS"), (address));
+        CURVE_USDT_POOL_ADDRESS = abi.decode(vm.parseJson(file, "CURVE_USDT_POOL_ADDRESS"), (address));
+        CURVE_COMPOUND_POOL_ADDRESS = abi.decode(vm.parseJson(file, "CURVE_COMPOUND_POOL_ADDRESS"), (address));
+        CURVE_Y_POOL_ADDRESS = abi.decode(vm.parseJson(file, "CURVE_Y_POOL_ADDRESS"), (address));
+        CURVE_3_POOL_ADDRESS = abi.decode(vm.parseJson(file, "CURVE_3_POOL_ADDRESS"), (address));
+        CURVE_TRICRYPTO2_POOL_ADDRESS = abi.decode(vm.parseJson(file, "CURVE_TRICRYPTO2_POOL_ADDRESS"), (address));
+        CURVE_ANKRETH_POOL_ADDRESS = abi.decode(vm.parseJson(file, "CURVE_ANKRETH_POOL_ADDRESS"), (address));
+        BALANCER_V2_ADDRESS = abi.decode(vm.parseJson(file, "BALANCER_V2_ADDRESS"), (address));
+
+        console2.logAddress(CURVE_COMPOUND_POOL_ADDRESS);
```
charlesjhongc, line 86: Remove?

test/utils/Addresses.sol
```diff
@@ -1,38 +1,105 @@
...
+        cDAI_ADDRESS = abi.decode(vm.parseJson(file, "cDAI_ADDRESS"), (address));
+        cUSDC_ADDRESS = abi.decode(vm.parseJson(file, "cUSDC_ADDRESS"), (address));
+        yDAI_ADDRESS = abi.decode(vm.parseJson(file, "yDAI_ADDRESS"), (address));
+        yUSDC_ADDRESS = abi.decode(vm.parseJson(file, "yUSDC_ADDRESS"), (address));
+        yUSDT_ADDRESS = abi.decode(vm.parseJson(file, "yUSDT_ADDRESS"), (address));
+        yTUSD_ADDRESS = abi.decode(vm.parseJson(file, "yTUSD_ADDRESS"), (address));
+        console2.logAddress(yUSDT_ADDRESS);
```
charlesjhongc, line 93: Remove?

charlesjhongc:
I'd prefer @pilagod's solution. I think read addresses from config file is good but having lots of setter doesn't seems worthy.

NIC619:
Updated with @pilagod 's suggestion.

test/RFQ.t.sol
```diff
@@ -58,9 +56,6 @@ contract RFQTest is StrategySharedSetup {
         // Setup
         if (!vm.envBool("DEPLOYED")) {
             // overwrite tokens with locally deployed mocks
```
charlesjhongc, line 58: Comment should be updated here.

charlesjhongc (approved):
LGTM. However there are lots of zero address in config of networks other than mainnet. This should be noticed since when running test in other network, some operation may failed. For example set ERC20 balance using `tokens` array. This can be adjusted in another PR though.

pilagod (approved):

PR #138, opened 2023-01-11
Update foundry and prettier ignore configs

NIC619:
Update
- foundry config
- prettier ignore config

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -3,7 +3,7 @@ src = 'contracts'                                             # the source direc
 test = 'test'                                                 # the test directory
 out = 'artifacts'                                             # the output directory (for artifacts)
 cache = true                                                  # whether to cache builds or not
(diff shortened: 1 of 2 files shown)


foundry.toml
```diff
@@ -3,7 +3,7 @@ src = 'contracts'                                             # the source direc
 test = 'test'                                                 # the test directory
 out = 'artifacts'                                             # the output directory (for artifacts)
 cache = true                                                  # whether to cache builds or not
-force = true                                                  # whether to ignore the cache (clean build)
+force = false                                                 # whether to ignore the cache (clean build)
```
NIC619, line 6: Foundry would recompile contracts/tests if they are modified so I think we can disable force build to save tens of seconds each time running tests locally

pilagod (approved):

charlesjhongc (approved):
The title and description could be imporoved.

PR #141, opened 2023-01-16
Apply two-step owership changing flow

charlesjhongc:
Applies to
- contracts/PermanentStorage.sol
- contracts/LimitOrder.sol
- contracts/UserProxy.sol

### DIFF
--- a/contracts/LimitOrder.sol
+++ b/contracts/LimitOrder.sol
@@ -38,6 +38,7 @@ contract LimitOrder is ILimitOrder, BaseLibEIP712, SignatureValidator, Reentranc
     address public coordinator;
     ISpender public spender;
     address public feeCollector;
+    address private nominatedOperator;
 
     // Factors
     uint16 public makerFeeFactor = 0;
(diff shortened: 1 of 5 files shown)


contracts/LimitOrder.sol
```diff
@@ -78,11 +79,19 @@ contract LimitOrder is ILimitOrder, BaseLibEIP712, SignatureValidator, Reentranc
...
-        emit TransferOwnership(_newOperator);
+        emit OperatorNominated(_newOperator);
+    }
+
+    function acceptOwnership() external {
+        require(msg.sender == nominatedOperator, "LimitOrder: not nominated");
+        emit OperatorChanged(operator, nominatedOperator);
+
+        operator = nominatedOperator;
+        nominatedOperator = address(0);
```
pilagod, line 94: Not sure if we need time lock here as well? Although splitting into two steps, if no time lock, user is still in the risky situation.

changwu-tw (approved):
Same as we did before

contracts/LimitOrder.sol
NIC619: I think this is mainly to prevent operator error.

NIC619 (approved):

contracts/LimitOrder.sol
charlesjhongc: Yeah, this prevent incorrect operator to be assigned.

PR #144, opened 2023-01-16
Cache LimitOrder storage var

charlesjhongc:
Add memory cache for `spender` and `feeCollector` vars.


pilagod (approved):
So it's worth, with regard to gas cost, caching variables in memory if they are referred >= 2 times?

changwu-tw (approved):

NIC619:
Maybe use `forge snapshot` to compare gas diff?

charlesjhongc:
Cache storage variable in memory is a common way in gas optimization. Let's merge this first then I'll try to compare the diff.

PR #145, opened 2023-01-16
Deprecate oz safe approve

charlesjhongc:
Replace `safeApprove()` with standard `approve()` function in LimitOrder. The approval is one-time use and should be zero after swap.

### DIFF
--- a/contracts/LimitOrder.sol
+++ b/contracts/LimitOrder.sol
@@ -431,7 +431,7 @@ contract LimitOrder is ILimitOrder, BaseLibEIP712, SignatureValidator, Reentranc
     }
 
     function _swapByProtocol(ProtocolSettlement memory _settlement) internal returns (uint256 amountOut) {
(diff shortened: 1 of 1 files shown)


pilagod:
I'm not sure removing 0 reset is good or not, since the approve front run issue is caused by remaining allowance. ([ref](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/18691d1a6f56dd906c175072a97dc5ed93ab968c/contracts/token/ERC20/IERC20.sol#L57-L62)). I'd like to know others' thoughts.

NIC619:
Approve 0 is used to make sure no allowance left for external contracts to be safe. And if one day we decide to add swap function like `exactOutput` (not `exactInput`), there could be allowance left.

And `safeApprove` is for non-standard ERC20.

changwu-tw:
Think that we can keep it unchanged?

charlesjhongc:
Yeah I think we can leave it unchanged.

PR #148, opened 2023-01-16
Optimize for loop with pre increment

charlesjhongc:
Ref : https://github.com/byterocket/c4-common-issues/blob/main/0-Gas-Optimizations.md/#g012---use-prefix-increment-instead-of-postfix-increment-if-possible


changwu-tw (approved):

pilagod (approved):
I feel that the gas cost between `i++` and `++i` is not so significant. To this case I would prefer to follow the way most developers used to code `i++`.

NIC619 (approved):

charlesjhongc:
@pilagod If the loop iterate many times then the difference may be worthy. I think it could be the best practice although this should be optimized in compiler I think.

PR #150, opened 2023-01-18
Check required address not zero in each tests

NIC619:
Check used addresses are not zero addresses and update used tokens during `setup` in each test.

### DIFF
--- a/test/forkMainnet/AMMQuoter.t.sol
+++ b/test/forkMainnet/AMMQuoter.t.sol
@@ -45,9 +45,10 @@ contract AMMQuoterTest is StrategySharedSetup {
         vm.label(address(this), "TestingContract");
         vm.label(address(ammQuoter), "AMMQuoterContract");
         vm.label(UNISWAP_V2_ADDRESS, "UniswapV2");
-        vm.label(SUSHISWAP_ADDRESS, "Sushiswap");
         vm.label(UNISWAP_V3_ADDRESS, "UniswapV3");
         vm.label(UNISWAP_V3_QUOTER_ADDRESS, "UniswapV3Quoter");
+        vm.label(SUSHISWAP_ADDRESS, "Sushiswap");
+        vm.label(BALANCER_V2_ADDRESS, "Balancer");
     }
 
     /*********************************
@@ -55,6 +56,16 @@ contract AMMQuoterTest is StrategySharedSetup {
      *********************************/
 
     function testSetupAMMQuoter() public {
+        // Check fork mainnet addresses are not zero addresses
+        assertTrue(uint160(UNISWAP_V2_ADDRESS) != 0);
+        assertTrue(uint160(UNISWAP_V3_ADDRESS) != 0);
+        assertTrue(uint160(UNISWAP_V3_QUOTER_ADDRESS) != 0);
+        assertTrue(uint160(SUSHISWAP_ADDRESS) != 0);
+        assertTrue(uint160(BALANCER_V2_ADDRESS) != 0);
+        assertTrue(uint160(WETH_ADDRESS) != 0);
+        assertTrue(uint160(DAI_ADDRESS) != 0);
+        assertTrue(uint160(USDT_ADDRESS) != 0);
+
         assertEq(address(ammQuoter.permStorage()), address(permanentStorage));
         assertEq(ammQuoter.weth(), WETH_ADDRESS);
     }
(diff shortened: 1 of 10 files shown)


test/forkMainnet/AMMQuoter.t.sol
```diff
@@ -45,16 +45,27 @@ contract AMMQuoterTest is StrategySharedSetup {
...
+        vm.label(BALANCER_V2_ADDRESS, "Balancer");
     }
 
     /*********************************
      *          Test: setup          *
      *********************************/
 
     function testSetupAMMQuoter() public {
+        // Check fork mainnet addresses are not zero addresses
+        assertTrue(uint160(UNISWAP_V2_ADDRESS) != 0);
```
pilagod, line 60: Q: Couldn't it compare to `address(0)` like `UNISWAP_V2_ADDRESS != address(0)`?

pilagod (approved):

charlesjhongc:
Since tests may be run in different network, zero check is also needed in tests not under `/forkMainnet`. I think maybe check zero address in `Addresses.sol` is much better since it's not relevant to each strategy contract.

test/forkMainnet/AMMQuoter.t.sol
NIC619: Somehow I mistaken an error as not able to compare but in fact it can. 🤦‍♂️

NIC619:
For not forking tests, all contracts should be freshly deployed so I think they don't need zero address check?

For checking zero address in `Addresses.sol`, that was my original plan but then I realized that not all contracts are going to be deployed on every network. We can check them on a per network basis but it will make the `Addresses.sol` look quite messy.
Also not all contracts are going to be needed by each test, so I ended up checking only the ones needed in each test.

pilagod:
Just raise a caveat here that most of non-forking tests indeed do not need addresses in the outside world, except for those tests depending on `BalanceUtil` like `SpenderTest`, since `BalanceUtil` depends on `Addresses` util.

In the long run it would be better to cut off this dependency, and `Addresses` should be only used in forking tests.

pilagod (approved):

NIC619:
Yeah this is a good point. I will update it in another PR.

charlesjhongc:
I agree that non-forking test should not depends on on-chain address. I think test cases should be separated in a better way so it can fit in different testing scenario including release test. But even for forked tests, I think leaving each test suite to assert non-zero addresses is against our design of environment abstraction. One solution is removing `Address.sol` from `BalanceUtil.sol` dependency so the zero address check can be moved to `Address.sol`.

NIC619:
I don't think we even have a design of environment abstraction as our tests basically only run on mainnet. If some contracts exist on network A but not exist on network B, you can not simply change the network of the tests that run fine on network A and run them on network B. There will be tests that can only run on certain network and so they check non-zero addresses for their use case in their test files.

My point is that environment abstraction does not make much sense because contracts exist/not-exist on different network, they are not simply just having different addresses. Even if you implement the environment abstraction by checking non-zero address in `Addresses.sol`, you still can't just run tests against different network.

pilagod:
I think it is not possible to check zero address in a single place since each network and test have their specific need. Require everything ahead might incur problem when we need to further scale our tests into more network environment. 

Or to keep it simple, I think is also ok to not check zero address at all, because the tests will fail and tell us when any required address is missing. Imo this is already enough to serve as a guard for us to the required address for tests.

charlesjhongc:
@NIC619 There are configs for different network under `test/utils/config/`. Isn't it an indication that tests may run in different environment/network? The idea of environment abstraction is that test suite should only depends on our contract implementation itself and should be independent from underlying environment. The deployed test is the first needs of environment abstraction and although tests are run in `forked mainnets (ethereum/arbitrum/...)` most of the time, deployed test needs to run in testnets as well after deploying. Also, from a long-term perspective, we should move toward to support multiple chains. That's the reason of separating test cases from underlying environmental setups and that's what I meant by `environment abstraction`.

However, it's also true that differences exist between networks. But I think instead of running test only in one network because of those differences, we should find out what those differences are and how to deal with it. Does our contract need to be changed to adapt different network? Or some cases should be filtered out for certain networks? I think these questions should be answered first then the rest should be technical details only. If just few tokens not exists on other network, we can specify network config to filter them out in `Address.sol` or other utils contract so they can be undefined in that situation.

I think we can move to offline discussion if needed.

ps. the current deployed mode could be further optimized. After all, running every test cases against deployed contract is a little more than ensuring contract is setup correctly.

NIC619:
Will go with not checking zero addresses and instead fill the address or skip test on that network when test errored.

I will open another PR to update the minor stuffs other than checking zero addresses in this PR.

PR #152, opened 2023-01-19
Implement new MMP contract

charlesjhongc:
Update MMP contract to be compliant with standard EIP-1271 interface and add unit test of MMP contract.


contracts/MarketMakerProxy.sol
```diff
@@ -4,102 +4,89 @@ pragma solidity 0.7.6;
...
     }
 
-    function withdraw(
+    function withdrawToken(
         address token,
-        address payable to,
+        address to,
         uint256 amount
-    ) public onlyWithdrawer {
+    ) external onlyOwner {
```
NIC619, line 77: I think this changes the usage pattern of MMP? If this is not customized on a per marker maker basis, maybe we need to think about if this suits their need?

Owner in my opinion would be a cold wallet/multi-sig, requiring it to execute a transaction each time to withdraw one asset might be a problem.

test/MarketMakerProxy.t.sol
```diff
@@ -0,0 +1,228 @@
...
+pragma solidity 0.7.6;
+pragma abicoder v2;
+
+import "@openzeppelin/contracts/math/SafeMath.sol";
+import "contracts/MarketMakerProxy.sol";
+import "test/utils/BalanceSnapshot.sol";
+import "test/utils/StrategySharedSetup.sol";
+import "test/mocks/MockERC1271Wallet.sol";
+
+contract MarketMakerProxyTest is StrategySharedSetup {
```
pilagod, line 11: MMP test seems not so relevant to other tokenlon infra contracts? Not sure if it is required to inherit `StrategySharedSetup` since it would cost to setup many other contracts that MMP test doesn't need.

contracts/MarketMakerProxy.sol
charlesjhongc: Right, I didn't know this consideration though. However if the `withdrawer` and `operator` are both EOA, what's the purpose to have a cold wallet or multisig owner since the allowance and the withdraw whitelist security is actually relied on EOA accounts?

test/MarketMakerProxy.t.sol
charlesjhongc: I was too lazy...fixed in a3804ee.

contracts/MarketMakerProxy.sol
```diff
@@ -4,102 +4,89 @@ pragma solidity 0.7.6;
...
 contract MarketMakerProxy is Ownable {
     using SafeERC20 for IERC20;
+    using Address for address payable;
 
-    address public SIGNER;
-    address public operator;
+    event ChangeSigner(address);
+    event UpdateWhitelist(address, bool);
+    event WrapETH(uint256);
+    event WithdrawETH(uint256);
```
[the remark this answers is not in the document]
charlesjhongc: Added in a3804ee.

contracts/MarketMakerProxy.sol
NIC619: Yeah, indeed the access control seems too simple to be effectively secure. If owner need to set withdrawer and also need to withdraw himself, then it doesn't seem like there's a need to set the withdrawer at all.🤔 

But if withdrawer can just withdraw all funds, then this extra layer of access control not help with the security at all, just like you point out.

I think there's two way forward, either we add in more fine-grain access control so adding withdrawers do help improve security or we just remove withdrawer completely. I prefer the latter.😂 MM can add in more fine-grain access control if they want.

charlesjhongc: Yeah actually I think this MMP should be just an example. The access control design is heavily depends on the need of each MM. I think the withdraw whitelist is still make sense even without withdrawer?

NIC619: Unless withdraw is called by non owner? Otherwise owner can just specify any `to` he likes since he is also the one controls whitelist?

pilagod (approved):

contracts/MarketMakerProxy.sol
charlesjhongc: Yeah since the owner controls everything. The whitelist here is kind of two step process to avoid typo or incorrect copy and paste. It's like the experience of CEX. Perhaps we can add a time lock in `updateWithdrawWhitelist()`?

NIC619: I think we can just leave it to MMs.

NIC619 (approved):

PR #157, opened 2023-02-06
Token Collector

pilagod:
Trial for a new role Token Collector, rationale document could see here:

https://imtoken.atlassian.net/wiki/spaces/IL/pages/1174864560/v6+Token+Collector


contracts/Spender.sol
```diff
@@ -190,6 +195,27 @@ contract Spender is ISpender, Ownable {
         _transferTokenFromUserTo(_user, _tokenAddr, _recipient, _amount);
     }
 
+    /// @dev Spend tokens on user's behalf with user's permit signature. Only an authority can call this.
+    /// @param _params The params of the SpendWithPermit.
+    /// @param _spendWithPermitSig Spend with permit signature.
+    function spendFromUserToWithPermit(SpenderLibEIP712.SpendWithPermit calldata _params, bytes calldata _spendWithPermitSig) external override onlyAuthorized {
```
pilagod, line 201: Add spend with permit for test and demo

contracts/TokenCollector.sol
```diff
@@ -0,0 +1,99 @@
...
+        address to,
+        uint256 amount,
+        bytes memory data
+    ) private {
+        if (data.length > 0) {
+            (SpenderLibEIP712.SpendWithPermit memory permit, ) = abi.decode(data, (SpenderLibEIP712.SpendWithPermit, bytes));
+            require(permit.tokenAddr == token, "TokenCollector: invalid token");
+            require(permit.user == from, "TokenCollector: invalid user");
+            require(permit.recipient == to, "TokenCollector: invalid recipient");
+            require(permit.amount == amount, "TokenCollector: invalid amount");
```
pilagod, line 75: For arbitrary data transfer we need to decode it to ensure the transfer data, like from, to and amount, are all as our expected.

pilagod: That's why for other sources would prefer to use permit + transferFrom pattern, since allowing arbitrary data for permit will not cause user loss, and the final transfer details are all guarded by us to zone out malicious actions.

package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "solhint \"contracts/**/*.sol\"",
     "compile": "forge build --force",
     "test-foundry-local": "DEPLOYED=false forge test --no-match-path 'test/forkMainnet/*.t.sol'",
-    "test-foundry-fork": "DEPLOYED=false forge test --fork-url $MAINNET_NODE_RPC_URL --fork-block-number 15451000 --match-path 'test/forkMainnet/*.t.sol'"
+    "test-foundry-fork": "DEPLOYED=false forge test --fork-url $MAINNET_NODE_RPC_URL --fork-block-number 15986406 --match-path 'test/forkMainnet/*.t.sol'"
```
pilagod, line 18: The block number permit2 deployed.

contracts/TokenCollector.sol
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    function collect(
+        address token,
+        address from,
+        address to,
+        uint256 amount,
+        bytes memory data
+    ) external override {
+        require(msg.sender == owner, "TokenCollector: not owner");
```
NIC619, line 34: If the owner (generic swap?) allows arbitrary message call, then it need to forbid arbitrary calls to this contract.

contracts/TokenCollector.sol
```diff
@@ -0,0 +1,99 @@
...
+
+    function collectFromToken(
+        address token,
+        address from,
+        address to,
+        uint256 amount,
+        bytes memory data
+    ) private {
+        if (data.length > 0) {
+            (bool success, ) = token.call(abi.encodePacked(IERC20Permit.permit.selector, data));
```
NIC619, line 57: The possibility that `data` can be completely irrelevant to `from`, `to`, `amount` does not sound good. I think we need to validate the params in `data` just like in `collectFromSpender`.

contracts/TokenCollector.sol
pilagod: Yes, for strategy supports arbitrary call it should responsible to block others to access token collector. Just like CowSwap does.

https://github.com/cowprotocol/contracts/blob/a7cb2900191c4871b72f2a8f3f81ba4029d14070/src/contracts/GPv2Settlement.sol#L456-L459

contracts/TokenCollector.sol
pilagod: For the highest security standard I agree that we should check all payloads are meet our expectation. Token seems a special case because its address is assigned by client, we might need more discussions.

But for like Uniswap Permit2, the address is saved on contract and is trusted by us, I think we could just let the permit go, if the data is wrong, then upcoming transfer would fail.

NIC619: If Alice signs permit but eventually the order was not settled on chain. Bob can take Alice's permit data on chain? (Bob has approve MAX already so token transfer from Bob will succeed, along with Alice's permit will succeed.) In this case, Alice's permit will be executed despite her order was not settled, though I haven't got a picture of how would this leads to exploit, this still does not seem like a desirable outcome.

doublespending: How to prevent reusing permitSign data in reverted transaction?

pilagod: Interesting question. I think this is a common permit problem I haven't thought before. When any contract supports permit, user must provide their permit sig for the contract to spend their tokens. But there is always a possibility that the tx finally reverts, and the permit sig is opened to everybody on chain.

Even we compose and check the payload by ourself, user still needs to provide permit sig along with those one-time fields (`nonce`, `expiry`) they signs in the calldata to us, and this can be easily replayed by any other seeing this tx.

NIC619: I was under the wrong impression that the contract accepting permit will check `msg.sender == spender`...
Then I guess this has to be guarded against by monitor or we just pray this wouldn't lead to any significant problem.😂 After all we are probably not the most profitable target to be exploited of this issue.

pilagod: Take permit2 as example, allowance permit indeed would take spender into account

https://github.com/Uniswap/permit2/blob/4382d768fb549af29f71e02e2c638fb3928d3b4b/src/interfaces/IAllowanceTransfer.sol#L54-L62

but for transfer permit it seems to allow anyone to transfer with permit

https://github.com/Uniswap/permit2/blob/4382d768fb549af29f71e02e2c638fb3928d3b4b/src/interfaces/ISignatureTransfer.sol#L19-L34

or maybe we should require user to also sign witness data like order hash to restrict the permit usage

https://github.com/Uniswap/permit2/blob/4382d768fb549af29f71e02e2c638fb3928d3b4b/src/interfaces/ISignatureTransfer.sol#L78-L95

but witness data doesn't seem robust enough too. If the tx is accidentally reverted, the transfer sig and data are still exposed to hacker, and they can directly transfer to token 😞

It's weird that permit2 doesn't even restrict the recipient in the permit. It looks like not so secure 🤨

NIC619: So it looks like for Permit2, we can check that `from` is the same as the `owner` in permit data to prevent stealing permit. But for native permit, we have to either monitor or just assume it's fine.

pilagod:
Here is the 1inch style implementation, which turns `TokenCollector` into an abstract contract and collect token would not need an additional external call.

https://github.com/consenlabs/tokenlon-contracts/compare/token-collector...token-collector-1inch

And the following are the gas reports for these two impls. To my surprise, the 1inch style seems to cost more than CowSwap style. Maybe it's because in the 1inch impl test there is an additional layer of strategy contract. I would further try to align two impls to the same condition.

* CowSwap style (a sidecar contract beside strategy contract (impl of this PR))
Screenshot 2023-02-14 at 4 15 49 PM

* 1inch style (an abstract contract inherited by strategy contract)
Screenshot 2023-02-14 at 4 08 23 PM

pilagod:
After the adjustment ([9f11747](https://github.com/consenlabs/tokenlon-contracts/pull/157/commits/9f11747803730230d9e1bbdf7b883ad8d98ea013)) to align the usage in these two cases, the CowSwap style indeed cost more gas

* CowSwap style
Screenshot 2023-02-15 at 2 55 05 PM

* 1Inch style
Screenshot 2023-02-14 at 4 08 23 PM

pilagod:
I have moved the implementation toward the 1inch style to acheive less gas consumption. Additional features like supporting both permit2 allowance transfer & signature transfer, we could do it in another PR. Merge this first could push other strategies design forward. 👍

charlesjhongc (commented):
Looks great. But are you gonna integrate allowance target and signature transfer of permit2 as well?

pilagod:
@charlesjhongc Yes, after this PR is merged, I will first add permi2 signature transfer. Allowance target part can be added after we decide how to implement it.

test/mocks/MockERC20Permit.sol
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts v4.4.1 (token/ERC20/extensions/IERC20Permit.sol)
+
+// NOTE: Internally downgrade to Solidity 0.7.6 and refine to fit our usage
```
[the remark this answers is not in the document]
pilagod: Nice catch, I totally forgot this 🤦‍♂️ I will switch to oz version.

test/forkMainnet/TokenCollector.t.sol
```diff
@@ -0,0 +1,109 @@
...
+    function setUp() public {
+        token.mint(user, 10000 * 1e18);
+
+        vm.label(address(this), "TestingContract");
+        vm.label(address(token), "TKN");
+    }
+
+    /* Token */
+
+    function testCollectByTokenApproval() public {
```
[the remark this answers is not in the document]
pilagod: I think strategies only need to care their own logic. I will add more edge tests to the Token Collector.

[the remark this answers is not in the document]
pilagod: This takes much more works than I expected... Patched in [271d266](https://github.com/consenlabs/tokenlon-contracts/pull/157/commits/271d266e8809257ba6454ea9bb8876bb586eff15)

contracts/abstracts/TokenCollector.sol
```diff
@@ -61,8 +66,11 @@ abstract contract TokenCollector {
     ) private {
         require(amount < uint256(type(uint160).max), "TokenCollector: permit2 amount too large");
         if (data.length > 0) {
-            (bool success, ) = uniswapPermit2.call(abi.encodePacked(IUniswapPermit2.permit.selector, data));
-            require(success, "TokenCollector: permit2 permit failed");
+            (address owner, IUniswapPermit2.PermitSingle memory permit, bytes memory permitSig) = abi.decode(
+                data,
+                (address, IUniswapPermit2.PermitSingle, bytes)
+            );
+            IUniswapPermit2(uniswapPermit2).permit(owner, permit, permitSig);
```
pilagod, line 73: Permit2 uses custom error. I didn't find a good way to keep the original error along the way to our client from the `result` returned by `call`, so I switched to explicitly call the `permit` function

```solidity
(bool success, bytes memory result) = uniswapPermit2.call(abi.encodePacked(IUniswapPermit2.permit.selector, data));
if (!success) {
    // how to revert?
}
```

charlesjhongc: I think we can refactor this in the later starge, let's focus on the functionality first.

charlesjhongc (approved):
LGTM

doublespending (approved):

pilagod:
@NIC619 Since there are already two approvals, I would merge this PR first for further integration. Feel free to comment new findings, I would patch them in the upcoming PR. 🙌

PR #158, opened 2023-02-07
`BalanceUtil` not inherit `Addresses`

NIC619:
Approach 1 (48edaee). `setERC20Balance` update `totalSupply` in local, not in fork
Approach 2 (85af8fa). try catch if `totalSupply` can be update regardless of fork

### DIFF
--- a/test/utils/BalanceUtil.sol
+++ b/test/utils/BalanceUtil.sol
@@ -7,12 +7,26 @@ import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 contract BalanceUtil is Test {
     using stdStorage for StdStorage;
 
+    function externalDeal(
+        address tokenAddr,
+        address userAddr,
+        uint256 amountInWei,
+        bool updateTotalSupply
+    ) public {
+        deal(tokenAddr, userAddr, amountInWei, updateTotalSupply);
+    }
+
     function setERC20Balance(
         address tokenAddr,
         address userAddr,
         uint256 amount
     ) internal {
         uint256 decimals = uint256(ERC20(tokenAddr).decimals());
-        deal(tokenAddr, userAddr, amount * (10**decimals), isFork() ? false : true);
+        uint256 amountInWei = amount * (10**decimals);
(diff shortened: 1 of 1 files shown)


NIC619:
Our test is having issue with Foundry's latest build so CI run failed. The tests ran successfully on my computer. You can take a look at the two approaches first.

pilagod:
Skipping supply update in fork env might cause unexpected state, right? (since new dealt tokens are not counted into the supply)

NIC619:
Main reason to update total supply is so that when doing normal transfers in test, it won't result in underflow when subtracting total supply. This used to be the case in our deployed WETH but since our mock WETH now inherits from ERC20 so we can now update WETH's total supply. Though we still need to [actually deposit](https://github.com/consenlabs/tokenlon-contracts/blob/06a178b52409ecd6e73abd7ee957ed452d7e77bd/test/RFQ.t.sol#L83-L84) so it have enough for withdrawal...

For fork network, since all of the tokens we use have enough total supply, they won't result in underflow problem. 🤣

NIC619:
For local, we are using mock ERC20. But using mock ERC20 in fork to replace the actual token contracts are not worth it in my opinion - losing the chance to interact with real token contracts only for the ability to update WETH's total supply.

NIC619:
I personally prefer try-catch approach as it has less mental complexity - it updates `balance`/`totalSupply` on tokens it can and only updates `balance` on token it can't, and the WETH token in fork env is really the only token that can't have `totalSupply` updated.

pilagod:
Me too for the try catch solution. It's more intuitive to me, and keep consistency between different environment could reduce unexpected outcomes.

charlesjhongc:
Sure, I mean **handling mocked ERC20 to have enough supply when deploying them in local tests**. Then there's no need to worry total supply in `BalanceUtil` at all.

NIC619:
@charlesjhongc @pilagod Let's go with try/catch approach?

pilagod (approved):

charlesjhongc (approved):
LGTM.

PR #162, opened 2023-02-13
[v6] Upgrade Solidity to 0.8.17

pilagod:
* Upgrade Solidity to 0.8.17 (foundry doesn't know 0.8.18 yet)
    * Utils (library, abstract etc.) uses `^0.8.0`
    * Interfaces uses `>=0.8.0`
* Upgrade `openzeppelin`, `forge-std` to latest 
* Categorize contracts into `abstracts`, `interfaces` and `libraries` for clarity (instead of a single big `utils`)
* Remove v5 files (contracts, docs etc.)
* Remove unused node packages


NIC619:
> Categorize contracts into abstracts, interfaces and libraries for clarity

I'm not sure `abstracts` category is going to help as we don't operate with the awareness of "which contract is an abstract contract". Or more broadly speaking, separating them into `abstracts`, `interfaces` and `libraries` carry meaning to a compiler but not so much to a developer so it does not help much in my opinion.

After looking at the contracts in `abstracts`, maybe we can just treat them as library (if not utils)? which means contracts in `libraries` do not necessary have to be `Library` type.

contracts/interfaces/IArbitrumL1Bridge.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity >=0.7.0;
+pragma solidity >=0.8.0;
```
charlesjhongc, line 2: Minor: Do we need arbitrum interface in v6?

charlesjhongc:
Originally a `Lib` prefix is added to library files in order to help importer to determine what type of imported contract. I think @pilagod is trying to separate them by path so no prefix is needed now.

pilagod:
I think abstract contract and library do have differences that developers must be aware. For abstract contract we can only use it by inheritance, while library is just like a namespace for a collection of utils.

It's clearer to me to separate them. When importing things from `abstracts` folder, we will know them should be inherited by contract at the beginning. And when importing things from `libraries`, we will know them could be used directly or augment other types.

And just like @charlesjhongc said, we can also get rid of the prefix/suffix like `BaseXXX` and `LibXXX`, which adds no much meaning to the util.

```solidity
import { A } from "contracts/abstracts/A"
import { B } from "contracts/libraries/B"

contract Strategy is A {
    using B for type;
    
    constructor() A() {}
    
    action() external {
        B.action()
    }
}
```

contracts/interfaces/IArbitrumL1Bridge.sol
pilagod: Just keep the interfaces for convenience. I'm ok to remove them for now util we need them again 😆

NIC619:
But this would require developers to remember which contract is abstract, which is not right? Because neither the name `EIP712`, `Multicall` or `SignatureValidator` strike me as an abstract contract (more like library). But if number of contracts is kept low then I guess it won't be too much of a problem.

pilagod:
Developer doesn't need to remember, they can just go through these two directories to see which function he needs (just like we go through single `utils` folder). For example, when he wants to use `Multicall`, he can find and see it in the `abstracts` folder, then he knows it is a abstract contract, and he has to inherit it before usage.

Current popular projects is also not favor of one single `utils` folder with prefix/suffix to distinguish files, rather they group utils by meaning or characteristic. For example, in [Uniswap](https://github.com/Uniswap/v3-periphery/tree/main/contracts) `libraries` only for libraries, and `base` folder for abstract contracts. Another example is [Openzeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/token/ERC20), they structure folders purely by meaning (`presets`, `extensions`), and without any file prefix/suffix.

To summarize, imo one big `utils` folder is not a good way to manage a growing project in the long run. A more meaningful structure would help developers to understand and develop in a clear way, and I personally prefer Uniswap structure more than Openzeppelin.

NIC619:
> Developer doesn't need to remember, they can just go through these two directories to see which function he needs (just like we go through single utils folder).
> To summarize, imo one big utils folder is not a good way to manage a growing project in the long run. A more meaningful structure would...

I think that's where I would disagree because the structure is not meaningful in my opinion. I definitely prefer not having one big folder but if today we separated files into different folders and I still have to "go through every folder to find the file I'm looking for" then the separation is doing the opposite of helping.

I don't have a better idea of how to categorize the files right now and it's not a big problem so let's go with this design (`abstract`/`libraries`) now and see how it plays out.

charlesjhongc:
I'm positive with this direction. At least I think having the `libraries` path make sense to me now. But for the `abstract` path, perhaps we can separate contracts by their function or subject instead of their type of contract? It can be adjusted later and I think this PR is good to go.

pilagod:
OK, let's go for current structure at this moment. Feel free to open PR to refine it if there is any good ideas 💪

pilagod:
@charlesjhongc @NIC619 I also remove L2 relevant interfaces for now. If we need them later, we can move them back from `v5` branch. Check if you guys have other questions, or we can be ready to merge it to start building our new v6 protocol 💪

NIC619 (approved):

charlesjhongc (approved):

PR #163, opened 2023-02-14
Try gas report

NIC619:
- Add gas report commands
- Add gas report to CI workflow

### DIFF
--- a/.github/workflows/gas-diff.yml
+++ b/.github/workflows/gas-diff.yml
@@ -1,0 +1,44 @@
+name: Report gas diff
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths:
+      - contracts/**
+      - test/**
+      - foundry.toml
+      - remappings.txt
+      - .github/workflows/gas-diff.yml
+
+jobs:
+  compare_gas_reports:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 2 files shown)


NIC619:
Since `master` does not have a gas report file yet, CI runs in this PR can not produce gas **diff** report, only gas report. It's outputted to summary in each CI run, for example [here](https://github.com/consenlabs/tokenlon-contracts/actions/runs/4170383205#summary-11319409954).

I also tried outputting gas diff (if any) using sticky comment, you can see an example [here](https://github.com/consenlabs/tokenlon-contracts/pull/165#issuecomment-1429107482). Though I'm not sure if we need gas diff report comment in every PR.

pilagod:
Cool PR!

I see in the sticky PR, the report contains even mock contracts we use in tests. Not sure if the `ignore` option could filter test contracts out 🧐

```yaml
      - name: Compare gas reports
        uses: Rubilmax/foundry-gas-diff@v3.13.3
        with:
          ignore: test/**/* # filter out gas reports from specific paths (test/ is included by default)
        id: gas_diff
```

charlesjhongc:
I think we can dump gas diff using sticky comment in every PR.

NIC619:
Ignored gas report for test folder contracts in 293a99d and merged in sticky comment PR.

charlesjhongc (approved):

pilagod (approved):

PR #169, opened 2023-02-21
Implement GenericSwap

charlesjhongc:
TODO:
- ~Allow user submitting order signature to relayer instead of call contract directly.~
- ~Integrate with TokenCollector~


contracts/GenericSwap.sol
```diff
@@ -0,0 +1,108 @@
...
+        address _inputToken = swapData.inputToken;
+        address _outputToken = swapData.outputToken;
+
+        if (_inputToken.isETH() && msg.value != swapData.inputAmount) revert InvalidMsgValue();
+
+        if (!_inputToken.isETH()) {
+            _collect(_inputToken, taker, address(strategy), swapData.inputAmount, swapData.inputData);
+        }
+
+        strategy.executeStrategy{ value: msg.value }(_inputToken, _outputToken, swapData.inputAmount, swapData.strategyData);
```
doublespending, line 81: Should wrap ETH in strategy  contract?

contracts/GenericSwap.sol
```diff
@@ -0,0 +1,108 @@
...
+        */
+
+    constructor(address _uniswapPermit2) TokenCollector(_uniswapPermit2) {}
+
+    /// @param strategy The strategy contract
+    /// @param swapData Swap data
+    /// @return returnAmount Output amount of the swap
+    function executeSwap(IStrategy strategy, GenericSwapData calldata swapData) external payable override returns (uint256 returnAmount) {
+        return _executeSwap(strategy, swapData, msg.sender);
+    }
```
pilagod, line 48: Q: Should we also check swap hash here to avoid duplicate swap?

contracts/GenericSwap.sol
```diff
@@ -0,0 +1,108 @@
...
+        require(!filledSwap[swapHash], "already filled");
+        if (!SignatureValidator.isValidSignature(taker, swapHash, takerSig)) revert InvalidSignature();
+        filledSwap[swapHash] = true;
+        return _executeSwap(strategy, swapData, taker);
+    }
+
+    function _executeSwap(
+        IStrategy strategy,
+        GenericSwapData calldata swapData,
+        address taker
```
pilagod, line 70: Q: This role is more like a `maker` (or `owner`), right? Since we collect the swapping token from him, he is the one who makes / owns the order, and the strategy is more like a taker to fulfill the order for him.

contracts/GenericSwap.sol
```diff
@@ -0,0 +1,108 @@
...
+    /// @param swapData Swap data
+    /// @return returnAmount Output amount of the swap
+    function executeSwap(
+        IStrategy strategy,
+        GenericSwapData calldata swapData,
+        address taker,
+        uint256 salt,
+        bytes calldata takerSig
+    ) external payable override returns (uint256 returnAmount) {
+        bytes32 swapHash = getEIP712Hash(_getGSDataHash(swapData, salt));
```
pilagod, line 60: The signature seems to include the strategy data, but I think users don't need to care about the swap details, they only declare an order, and Tokenlon will somehow fulfill for them. 

Require users to sign swap routes beforehand might also restrict the SOR ability.

contracts/libraries/GeneralAsset.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { Address } from "@openzeppelin/contracts/utils/Address.sol";
+
+import { Constant } from "./Constant.sol";
+
+library GeneralAsset {
```
pilagod, line 10: nit: I personally would prefer to call it just `Asset`.   `General` doesn't bring additional meanings to me for this library. 😆

And it would be better to remove the function prefix also, and try to figure out more appropriate names for them. 

For instance:

* `generalBalanceOf` -> `getBalance`
* `generalTransfer` -> `transferTo`

contracts/GenericSwap.sol
charlesjhongc: It's left for strategy contract handling this since for some amm protocols, raw ETH is supported.

contracts/interfaces/IGenericSwap.sol
```diff
@@ -0,0 +1,35 @@
...
+    struct GenericSwapData {
+        address inputToken;
+        address outputToken;
+        uint256 inputAmount;
+        uint256 minOutputAmount;
+        address payable receiver;
+        uint256 deadline;
+        bytes inputData;
+        bytes strategyData;
+    }
```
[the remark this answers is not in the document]
charlesjhongc: I think the abstraction is about separating AMM interaction from the order. But we can discuss about having an uniform order struct in v6.

contracts/GenericSwap.sol
charlesjhongc: Since the caller here is the taker, if another identical swap is submitted then it means the taker wants to swap again?

pilagod: Yap, it is indeed a possible scenario 🧐 I think this also incurs another question, do we need to distinguish each order or not?

contracts/libraries/GeneralAsset.sol
charlesjhongc: I'm ok with the naming. It can be adjusted later as well.
Fixed in 59d2abfdee82605b53e91593a18c9a77a217afd2

contracts/GenericSwap.sol
pilagod: @doublespending You can check if you have any other inputs

doublespending (dismissed):

pilagod (approved):

contracts/GenericSwap.sol
charlesjhongc: After discussion, decided to include strategy data into signature for now.

contracts/GenericSwap.sol
charlesjhongc: Since the quote is provided by AMM protocols (or RFQ/LimitOrder), I still think it's considered as `taker` here. The taker is the one that accept the quote, which is given by maker.

contracts/GenericSwap.sol
charlesjhongc: I think we should prevent signature replay attack, but allow user to fill the same order multiple time by themself?

contracts/interfaces/IGenericSwap.sol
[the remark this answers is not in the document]
charlesjhongc: After discussion, have a shared `GeneralOrder` struct for now.

contracts/interfaces/IGeneralOrder.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+struct GeneralOrder {
```
pilagod, line 4: nit: It's general enough to call it `Order`.  Other specific type order (if there is in some future 🤣) could then use prefix/suffix to express, like `LimitOrder`, `SpecialOrder` etc.

contracts/interfaces/IGeneralOrder.sol
```diff
@@ -0,0 +1,17 @@
...
+    address payable maker;
+    address taker;
+    address inputToken;
+    bytes inputTokenPermit;
+    address outputToken;
+    bytes ourputTokenPermit;
```
pilagod, line 10:
```suggestion
    bytes outputTokenPermit;
```

pilagod (dismissed):

contracts/interfaces/IGeneralOrder.sol
charlesjhongc: I'll rename this part and update type hash in another PR.

pilagod (approved):

contracts/GenericSwap.sol
```diff
@@ -0,0 +1,77 @@
...
+pragma solidity 0.8.17;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import { TokenCollector } from "./abstracts/TokenCollector.sol";
+import { EIP712 } from "./abstracts/EIP712.sol";
+import { IGenericSwap } from "./interfaces/IGenericSwap.sol";
+import { IStrategy } from "./interfaces/IStrategy.sol";
+import { Order, getOrderHash, ORDER_TYPESTRING } from "./libraries/Order.sol";
```
pilagod, line 11:
```suggestion
import { Order, getOrderHash } from "./libraries/Order.sol";
```

contracts/libraries/Order.sol
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+bytes32 constant ORDER_DATA_TYPEHASH = 0xc504a4b52a554e816ce9776f48b79578cb4662dfb598d100bffef2eea538c5a5;
+
+string constant ORDER_TYPESTRING = "Order(address maker,address taker,address inputToken,bytes inputTokenPermit,address outputToken,bytes outputTokenPermit,uint256 inputAmount,uint256 outputAmount,uint256 minOutputAmount,address recipient,uint256 expiry,uint256 salt";
```
pilagod, line 6: Missing the end parentheses 

```suggestion
string constant ORDER_TYPESTRING = "Order(address maker,address taker,address inputToken,bytes inputTokenPermit,address outputToken,bytes outputTokenPermit,uint256 inputAmount,uint256 outputAmount,uint256 minOutputAmount,address recipient,uint256 expiry,uint256 salt)";
```

contracts/GenericSwap.sol
charlesjhongc: Fixed in 4945c5c.

contracts/libraries/Order.sol
charlesjhongc: Fixed in 4945c5c.

pilagod (approved):

doublespending (approved):

PR #170, opened 2023-02-22
Implement ammwrapper [approach 1]

doublespending:


contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,115 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+// interface, abstract, contract 0.8.17
+pragma abicoder v2;
+
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
```
charlesjhongc, line 6: `SafeMath` is not required in solidity ^0.8.0.

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,115 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+// interface, abstract, contract 0.8.17
+pragma abicoder v2;
```
charlesjhongc, line 4: Remove

charlesjhongc:
User `import { xxx } from "./XXX.sol` import pattern.

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,115 @@
...
+
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
```
charlesjhongc, line 10: User tokenlon `Ownable` contract instead.

contracts/interfaces/IAMMStrategy.sol
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+pragma abicoder v2;
```
charlesjhongc, line 3: Remove

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,115 @@
...
+
+import "./interfaces/IAMMStrategy.sol";
+
+contract AMMStrategy is IAMMStrategy, ReentrancyGuard, Ownable {
+    using SafeERC20 for IERC20;
+    using SafeMath for uint256;
+
+    address public entryPoint;
```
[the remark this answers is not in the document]
pilagod: If we go this way, then `setEntryPoint` could be removed together.

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,115 @@
...
+            require(ammMapping[op.dest], "invalid op dest");
+            opDests[i] = op.dest;
+            _call(op.dest, 0, op.data);
+        }
+        uint256 receivedAmount = IERC20(outputToken).balanceOf(address(this));
+        if (receivedAmount != 0) {
+            IERC20(outputToken).safeTransfer(entryPoint, receivedAmount);
+        }
+        uint256 balanceAfter = IERC20(outputToken).balanceOf(entryPoint);
+        emit Swapped(inputToken, inputAmount, opDests, outputToken, balanceAfter.sub(balanceBefore));
```
pilagod, line 96: I think maybe we could have a event for each operation, like cowswap

https://github.com/cowprotocol/contracts/blob/7efd5b43fbf46fe40bac9e6de1b6bab05d27c20b/src/contracts/GPv2Settlement.sol#L462-L466

This can log more info for each operation, and `opDests` could be removed in the swapped event.

test/forkMainnet/AMMStrategy/Setup.t.sol
```diff
@@ -0,0 +1,151 @@
...
+        bytes memory payload = abi.encodeCall(IUniswapV3SwapRouter.exactInputSingle, params);
+        IAMMStrategy.Operation[] memory operations = new IAMMStrategy.Operation[](1);
+        operation = IAMMStrategy.Operation(entry.makerAddr, payload);
+        operations[0] = operation;
+
+        data = abi.encode(operations);
+    }
+
+    function _sendTakerAssetFromEntryPoint(address takerAssetAddr, uint256 takerAssetAmount) internal {
+        vm.prank(entryPoint);
```
pilagod, line 147:
```suggestion
        vm.startPrank(entryPoint);
```

test/forkMainnet/AMMStrategy/TradeWithMultiAMM.t.sol
```diff
@@ -0,0 +1,55 @@
...
+        AMMStrategyEntry memory entryForV3 = DEFAULT_ENTRY;
+        entryForV3.makerAddr = UNISWAP_V3_ADDRESS;
+        entryForV3.takerAssetAmount = totalAmount - totalAmount / 2;
+
+        assertEq(entry.takerAssetAmount, entryForV2.takerAssetAmount + entryForV3.takerAssetAmount);
+
+        BalanceSnapshot.Snapshot memory entryPointTakerAsset = BalanceSnapshot.take(address(entryPoint), entry.takerAssetAddr);
+        BalanceSnapshot.Snapshot memory ammStrategyTakerAsset = BalanceSnapshot.take(address(ammStrategy), entry.takerAssetAddr);
+
+        // entryPoint sends taker asset to ammStrategy befor calling executeStrategy
```
pilagod, line 27:
```suggestion
        // entryPoint sends taker asset to ammStrategy before calling executeStrategy
```

test/forkMainnet/AMMStrategy/TradeWithMultiAMM.t.sol
```diff
@@ -0,0 +1,55 @@
...
+
+        operations[0] = operation1;
+        operations[1] = operation2;
+
+        entryPointTakerAsset = BalanceSnapshot.take(address(entryPoint), entry.takerAssetAddr);
+        ammStrategyTakerAsset = BalanceSnapshot.take(address(ammStrategy), entry.takerAssetAddr);
+        BalanceSnapshot.Snapshot memory entryPointMakerAsset = BalanceSnapshot.take(address(entryPoint), entry.makerAssetAddr);
+        BalanceSnapshot.Snapshot memory ammStrategyMakerAsset = BalanceSnapshot.take(address(ammStrategy), entry.makerAssetAddr);
+
+        vm.prank(entryPoint);
```
pilagod, line 46: nit: Maybe we could use the test contract itself as the entry point, and then we can directly call strategy without pranking

pilagod (commented):
nit: I think the tasks of this strategy are about (1) forward operation and calldata to external (2) ETH auto wrap/unwrap. So in my opinion the tests can be reduce to, for example, one multi AMM case (maybe 3 AMM uni, sushi, curve), two ETH as input/output cases, which are 3 cases in total.

Though testing each protocol separately can increase the confidence somehow, it would have more maintenance cost and doesn't focus enough on the core logic of this strategy.

Just my personal opinion, you can treat it as a minor reference. 😄

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+    modifier onlyEntryPoint() {
+        require(msg.sender == entryPoint, "only entry point");
+        _;
+    }
+
+    /************************************************************
+     *           Management functions for Owner               *
+     *************************************************************/
+    /// @inheritdoc IAMMStrategy
+    function setAMMs(address[] calldata _ammAddrs, bool[] calldata _enables) external override onlyOwner {
```
doublespending, line 57: * AMM cannot be an ERC-20 tokens, which user can change allowance of the tokens.
* For example, if WETH is an AMM, then user can `WETH.approve(SUSHISWAP,0)`

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+        address outputToken,
+        uint256 inputAmount,
+        bytes calldata data
+    ) external payable override onlyEntryPoint {
+        Operation[] memory ops = abi.decode(data, (Operation[]));
+        require(ops.length > 0, "empty operations");
+        require(inputAmount > 0, "empty inputAmount");
+        uint256 balanceBefore = Asset.getBalance(outputToken, entryPoint);
+        // wrap eth first
+        if (Asset.isETH(inputToken)) {
```
doublespending, line 109: * Wrap all native ETH first
* If operation need to use native ETH, native ETH will be withdraw in `_call` function

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+    function _call(
+        address _dest,
+        uint256 _value,
+        bytes memory _data
+    ) internal returns (bytes4 selector) {
+        if (_data.length >= 4) {
+            selector = bytes4(_data);
+        }
+        // withdraw needed native eth
+        if (_value > 0 && address(this).balance < _value) {
```
doublespending, line 136: Withdraw native ETH here if needed

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+    function approveTokens(
+        address[] calldata tokens,
+        address[] calldata spenders,
+        bool[] calldata usePermit2InSpenders,
+        uint256 amount
+    ) external override onlyOwner {
+        require(spenders.length == usePermit2InSpenders.length, "length of spenders not match");
+        for (uint256 i = 0; i < tokens.length; ++i) {
+            for (uint256 j = 0; j < spenders.length; ++j) {
+                if (usePermit2InSpenders[j]) {
```
doublespending, line 74: Uniswap universal router must use Permit2

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * @dev internal function of `executeStrategy`.
+     * Transfer output token to entry point
+     * If outputToken is native ETH and there is WETH remain, unwrap WETH to ETH automatically
+     */
+    function _transferToEntryPoint(address _token) internal {
+        if (Asset.isETH(_token)) {
+            // unwrap existing WETH
+            uint256 wethBalance = IWETH(weth).balanceOf(address(this));
```
doublespending, line 171: * unwrap WETH if outputToken is native ETH

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+                    _permit2Approve(tokens[i], spenders[j], amount);
+                } else {
+                    IERC20(tokens[i]).safeApprove(spenders[j], amount);
+                }
+            }
+        }
+    }
+
+    /// @inheritdoc IAMMStrategy
+    function withdrawLegacyTokensTo(address[] calldata tokens, address receiver) external override onlyOwner {
```
doublespending, line 85: There may be some legacy tokens after swap.

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+                revert(add(result, 32), mload(result))
+            }
+        }
+    }
+
+    /**
+     * @dev internal function of `executeStrategy`.
+     * Allow the spender to use Permit2 for the token.
+     */
+    function _permit2Approve(
```
doublespending, line 151: We should first approve permit2 and then approve in permit2

contracts/AMMStrategy.sol
pilagod: Treat WETH as an AMM is a cool idea that I never thought before 👍

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { Ownable } from "./abstracts/Ownable.sol";
+import { Asset } from "./libraries/Asset.sol";
+
+import { IWETH } from "./interfaces/IWeth.sol";
+import { IUniswapPermit2 } from "./interfaces/IUniswapPermit2.sol";
+import { IAMMStrategy } from "./interfaces/IAMMStrategy.sol";
+import { IStrategy } from "./interfaces/IStrategy.sol";
```
charlesjhongc, line 13: I think `IStrategy` can be removed here and @inheritdoc using `IAMMStrategy`.

contracts/AMMStrategy.sol
pilagod: Q: So universal router will use plain `transferFrom` from permit2, without any permit signature?

pilagod (approved):

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+    function executeStrategy(
+        address inputToken,
+        address outputToken,
+        uint256 inputAmount,
+        bytes calldata data
+    ) external payable override onlyEntryPoint {
+        Operation[] memory ops = abi.decode(data, (Operation[]));
+        require(ops.length > 0, "empty operations");
+        require(inputAmount > 0, "empty inputAmount");
+        uint256 balanceBefore = Asset.getBalance(outputToken, entryPoint);
```
charlesjhongc, line 107: Since all of the balance will be transferred before return, there's no need to calculate before/after balance diff?

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+            bytes4 selector = _call(op.dest, op.value, op.data);
+            emit Action(op.dest, op.value, selector);
+        }
+        _transferToEntryPoint(outputToken);
+        uint256 balanceAfter = Asset.getBalance(outputToken, entryPoint);
+        emit Swapped(inputToken, inputAmount, outputToken, balanceAfter - balanceBefore);
+    }
+
+    /**
+     * @dev internal function of `approveTokens`.
```
charlesjhongc, line 124: incorrect

contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+        require(ops.length > 0, "empty operations");
+        require(inputAmount > 0, "empty inputAmount");
+        uint256 balanceBefore = Asset.getBalance(outputToken, entryPoint);
+        // wrap eth first
+        if (Asset.isETH(inputToken)) {
+            IWETH(weth).deposit{ value: inputAmount }();
+        }
+        for (uint256 i = 0; i < ops.length; ++i) {
+            Operation memory op = ops[i];
+            require(ammMapping[op.dest], "invalid op dest");
```
charlesjhongc, line 114: Perhaps this check and event can be done within `_call()` function.

contracts/libraries/UniswapCommands.sol
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.17;
+
+/// @title Commands
+/// @notice Command Flags used to decode commands
+library Commands {
```
charlesjhongc, line 6: If this is only used in tests, move it under `test/libraries` instead.

test/forkMainnet/AMMStrategy.t.sol
```diff
@@ -0,0 +1,404 @@
...
+        } else {
+            IERC20(inputToken).safeTransfer(address(ammStrategy), inputAmount);
+            IStrategy(ammStrategy).executeStrategy(inputToken, outputToken, inputAmount, data);
+        }
+
+        inputTokenBalance.assertChange(-int256(inputAmount));
+        outputTokenBalance.assertChangeGt(0);
+    }
+
+    function setBalance(
```
charlesjhongc, line 376: User `setEOABalanceAndApprove()` and `deal()` instead. See `test/utils/BalanceUtil.sol`

contracts/AMMStrategy.sol
charlesjhongc: Yes, universal router utilize allowance transfer instead of signature transfer.

doublespending: Note: The UniversalRouter uses Permit2 to remove the need for token approvals being provided directly to the UniversalRouter. The Permit2 documentation can be found [here](https://docs.uniswap.org/contracts/permit2/overview).

I think so in command like `V2_SWAP_EXACT_IN`
* `address` The recipient of the output of the trade
* `uint256` The amount of input tokens for the trade
* `uint256` The minimum amount of output tokens the user wants
* `address[]` The UniswapV2 token path to trade along
* `bool` A flag for whether the input tokens should come from the msg.sender (through Permit2) or whether the funds are already in the UniversalRouter

contracts/AMMStrategy.sol
doublespending: It is need just when `emit Swapped(inputToken, inputAmount, outputToken, balanceAfter - balanceBefore);`
If we don't emit the Swap event, It's ok.

contracts/AMMStrategy.sol
doublespending: There will a compiler error: 
```
Compiler run failed
error[4682]: DocstringParsingError: Documentation tag @inheritdoc references contract "IAMMStrategy", but the contract does not contain a function that is overridden by this function.
  --> contracts/AMMStrategy.sol:97:5:
   |
97 |     /// @inheritdoc IAMMStrategy
```

contracts/AMMStrategy.sol
charlesjhongc: The `balanceAfter - balanceBefore` here is equal to `erc20.balanceOf(address(this))`.

contracts/AMMStrategy.sol
charlesjhongc: 8cff641

contracts/AMMStrategy.sol
charlesjhongc: 8cff641

contracts/AMMStrategy.sol
charlesjhongc: After discussion, remove `Swapped` event here and leave it to `GenericSwap`.

test/forkMainnet/AMMStrategy.t.sol
charlesjhongc: 6a2cdef

contracts/libraries/UniswapCommands.sol
charlesjhongc: 6a2cdef

charlesjhongc (approved):

pilagod (dismissed):

charlesjhongc (approved):

pilagod (approved):

PR #171, opened 2023-02-28
Implement ammwrapper [approach 2]

doublespending:


contracts/AMMStrategy.sol
```diff
@@ -0,0 +1,270 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { SafeMath } from "@openzeppelin/contracts/utils/math/SafeMath.sol";
```
charlesjhongc, line 6: `SafeMath` is not needed when using `^0.8.0`, can be removed.

pilagod:
@charlesjhongc @doublespending 

I feel that `AMMStrategy` in #170 is well and flexible enough to cover all the features in this strategy. Not sure if I miss something that in what circumstance it would be better for client to use this hardcoding version?

doublespending:
> 



> @charlesjhongc @doublespending
> 
> I feel that `AMMStrategy` in #170 is well and flexible enough to cover all the features in this strategy. Not sure if I miss something that in what circumstance it would be better for client to use this hardcoding version?

I think so and it's more flexiable, but it looks like that implemetation  in #170 will leave all dirty work to front-end. In addition, we are not sure is there any unaware attack vector in #170.

pilagod:
I think the amm whitelist in #170 is a good design to reduce the attack vector. (though I also cannot 100% promise there is no other problem 😂)

Another concern of hardcoding impl is that it will open up our own data interface, and client needs to learn the interface first in order to integrate our strategy. In contrast, the AMM protocols (uniswap, curve, ...) already provide a robust and well-known interface that most developers are familiar, #170 would reduce their mental cost to integrate our strategy.

And I think either this or #170 the frontend effort would be similar 😆

doublespending:
> I think the amm whitelist in #170 is a good design to reduce the attack vector. (though I also cannot 100% promise there is no other problem 😂)
> 
> Another concern of hardcoding impl is that it will open up our own data interface, and client needs to learn the interface first in order to integrate our strategy. In contrast, the AMM protocols (uniswap, curve, ...) already provide a robust and well-known interface that most developers are familiar, #170 would reduce their mental cost to use our strategy.
> 
> And I think both this and #170 the frontend effort would be similar 😆

Approach 2 has to integrate amms  one by one and it means that we have to update our contract and integrate uniswap  v4 &  curve v3 & .... We can save a lot of time and do other interesting  
 things in Approach 1.

PR #173, opened 2023-03-02
[v6] Support permit2 signature transfer in TokenCollector

pilagod:
Add permit2 signature transfer support in TokenCollector

### DIFF
--- a/contracts/abstracts/TokenCollector.sol
+++ b/contracts/abstracts/TokenCollector.sol
@@ -12,13 +12,13 @@ abstract contract TokenCollector {
 
     enum Source {
         Token,
-        UniswapPermit2
+        Permit2AllowanceTransfer
     }
 
-    address public immutable uniswapPermit2;
+    address public immutable permit2;
 
-    constructor(address _uniswapPermit2) {
-        uniswapPermit2 = _uniswapPermit2;
+    constructor(address _permit2) {
+        permit2 = _permit2;
     }
 
     function _collect(
@@ -32,8 +32,8 @@ abstract contract TokenCollector {
         if (src == Source.Token) {
             return _collectFromToken(token, from, to, amount, srcData);
         }
-        if (src == Source.UniswapPermit2) {
-            return _collectFromUniswapPermit2(token, from, to, amount, srcData);
+        if (src == Source.Permit2AllowanceTransfer) {
+            return _collectFromPermit2AllownaceTransfer(token, from, to, amount, srcData);
         }
         revert("TokenCollector: unknown token source");
     }
@@ -57,7 +57,7 @@ abstract contract TokenCollector {
         IERC20(token).safeTransferFrom(from, to, amount);
     }
 
-    function _collectFromUniswapPermit2(
+    function _collectFromPermit2AllownaceTransfer(
         address token,
         address from,
         address to,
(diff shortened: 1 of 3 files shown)


contracts/abstracts/TokenCollector.sol
```diff
@@ -70,8 +74,25 @@ abstract contract TokenCollector {
...
+        address from,
+        address to,
+        uint256 amount,
+        bytes memory data
+    ) private {
+        require(data.length > 0, "TokenCollector: permit2 data cannot be empty");
+        (IUniswapPermit2.PermitTransferFrom memory permit, IUniswapPermit2.SignatureTransferDetails memory detail, address owner, bytes memory permitSig) = abi
+            .decode(data, (IUniswapPermit2.PermitTransferFrom, IUniswapPermit2.SignatureTransferDetails, address, bytes));
+        require(permit.permitted.token == token, "TokenCollector: token address doesn't match");
+        require(owner == from, "TokenCollector: from address doesn't match");
```
charlesjhongc, line 93: What's the purpose to have `owner` encoded in data? I think maybe this check can be left to `permitTransferFrom()` of Permit2 by brining `from` address directly to the function.

pilagod: I agree that directly calling permit2 with `from` as owner is a good way, but this would cause client to encode data by our data format instead of permit2's interface.

I would prefer to let clients stick to permit2 native interface, so they don't need to learn additional specs in order to transfer tokens. 

Would love to hear more thoughts from others.

charlesjhongc: But encoding data with our format is inevitable and it could also reduce the length of calldata. The user may use permit2 SDK so having the same format may avoid re-encoding, but the re-encoding can save ~30 bytes in calldata which I think is worthy.

pilagod: I do the gas comparison and here is the result: (only run the success case to rule out noises)

* Current
before

* Custom format
after

Custom format implementation could refer to https://github.com/consenlabs/tokenlon-contracts/compare/permit2-sig-transfer...permit2-gas-diff

To summarize the changes in custom format version:

1. Gas consumption reduces ~1000.
2. The client still needs to sign permit2 struct (`IUniswapPermit2.PermitTransferFrom`) to create valid permit signature. 
    - There is an implicit rule in the struct that the permitted amount should be equal to requested amount.
    - Client has to extract `nonce`, `deadline` from permit2 struct, and combine them with permit signature, to encode the data to us.

We can discuss about which version would you guys prefer. After reaching a consensus, I will go for that direction.

cc. @charlesjhongc @doublespending @NIC619

charlesjhongc: I still put gas consumption at the higher priority so I think only required values should be brought as parameters.

pilagod: Refactor in [db5b60c](https://github.com/consenlabs/tokenlon-contracts/pull/173/commits/db5b60c26d174d72f9f3608c95f84a35fb6f1d33), also refactor allowance transfer in [367767b](https://github.com/consenlabs/tokenlon-contracts/pull/173/commits/367767b43cd1e2d4d4a8b617ddfa1830667b66b5)

charlesjhongc (approved):
LGTM

doublespending (approved):

contracts/abstracts/TokenCollector.sol
doublespending: I prefer to bring values as parameters, for the input parameters of `PermitSingle` are fixed and need not to let them free. Also, we should make sure that the encoded `data` and the inputs of `_collectByPermit2AllownaceTransfer` are match.

PR #187, opened 2023-05-08
RFQv2 block flashloan and unwrap maker WETH

charlesjhongc:


pilagod (commented):
I found we might miss one test case whose context is that, taker token is `WETH` and test maker receive `ETH`

charlesjhongc:
@pilagod Yeah right, it's for the case that I added in 7c3b247 and I forgot to add test for that. Added in 0da55a7.

pilagod (approved):

test/utils/Addresses.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity 0.7.6;
 import "forge-std/Test.sol";
 
 address constant ETH_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+address constant ZERO_ADDRESS = address(0);
```
NIC619, line 7: Is zero address a new convention to refer to ETH? Why do we need it?

charlesjhongc: Actually it's not new and it's aligned with `AMMWrapper` contract.
https://github.com/consenlabs/tokenlon-contracts/blob/v5/contracts/AMMWrapper.sol#L161

NIC619 (approved):

PR #190, opened 2023-05-14
RFQ add allowContract and unwrap weth if possible

charlesjhongc:
- Unwrap to ETH if takerToken == WETH
- Add `allowContractCall` to allow contract interaction
- Add more tests


test/forkMainnet/RFQ.t.sol
```diff
@@ -6,14 +6,16 @@ import { Tokens } from "test/utils/Tokens.sol";
...
 import { RFQ } from "contracts/RFQ.sol";
 import { IRFQ } from "contracts/interfaces/IRFQ.sol";
-import { IWETH } from "contracts/interfaces/IWeth.sol";
+import { IWETH } from "contracts/interfaces/IWETH.sol";
 import { TokenCollector } from "contracts/abstracts/TokenCollector.sol";
 import { Offer, getOfferHash } from "contracts/libraries/Offer.sol";
 import { RFQOrder, getRFQOrderHash } from "contracts/libraries/RFQOrder.sol";
 import { Constant } from "contracts/libraries/Constant.sol";
 
+// FIXME add taker = eoa + spender & maker = contract + directly approval
```
NIC619, line 18: This will be added in following PR?

contracts/libraries/Offer.sol
```diff
@@ -14,6 +14,7 @@ struct Offer {
     address makerToken;
     uint256 makerTokenAmount;
     uint256 minMakerTokenAmount;
+    bool allowContractCall;
```
NIC619, line 17: Maybe something like `allowContractSender` to avoid confusing it with "contract call"?

test/forkMainnet/RFQ.t.sol
charlesjhongc: Yeah will do after spender is added.

contracts/libraries/Offer.sol
charlesjhongc: I'm ok with the naming.

pilagod: I'm ok with both.

charlesjhongc: Renamed fd1d45d

NIC619 (approved):

pilagod (approved):

PR #192, opened 2023-05-16
add v6 Spender contract

charlesjhongc:


pilagod:
I remembered in the past discussion we decide to get rid of spender in v6. Not sure if I missed something and would like to know the purpose behind it. 🙌

charlesjhongc:
Need to rename so close and open another PR.

PR #196, opened 2023-05-17
Add v6 AllownaceTarget

charlesjhongc:


test/Spender.t.sol
```diff
@@ -0,0 +1,105 @@
...
+
+        uint256 amount = 100;
+        vm.prank(authorized);
+        allowanceTarget.spendFromUserTo(user, address(mockERC20), recipient, amount);
+
+        fromBalance.assertChange(-int256(amount));
+        toBalance.assertChange(int256(amount));
+    }
+
+    function testSpendFromUserToWithNoReturnValueToken() public {
```
pilagod, line 81: I think we should stick to specific test naming rule. I saw some cases above use underline `...Balance_ReturnFalseToken` but here uses `...WithNoReturnValueToken`. 

I would personally prefer no underline, but both are ok to me, as long as they are in an uniform style.

charlesjhongc: let's drop the underline.

NIC619 (approved):

pilagod (approved):

PR #197, opened 2023-05-18
SignalBuy contract

NIC619:
- Modified from LimitOrder contracts
- It checks takerToken/makerToken ratio and make sure the ratio provided by taker is better than or equal to order's specified ratio
- Since maker can receive more taker token if takerToken/makerToken ratio is better, the token type of the filled token amount record is changed from taker token to maker token.
    - added `orderHashToMakerTokenFilledAmount` to `LibOrderStorage`
- add tests to check
    - worse takerToken/makerToken ratio can not be filled and 
    - better takerToken/makerToken ratio can be filled fully/multiple times

### DIFF
--- a/contracts/SignalBuyContract.sol
+++ b/contracts/SignalBuyContract.sol
@@ -2,21 +2,21 @@
 pragma solidity 0.7.6;
 pragma abicoder v2;
 
-import "@openzeppelin/contracts/math/Math.sol";
-import "@openzeppelin/contracts/math/SafeMath.sol";
-import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
-import "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+import { Math } from "@openzeppelin/contracts/math/Math.sol";
+import { SafeMath } from "@openzeppelin/contracts/math/SafeMath.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+import { ReentrancyGuard } from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
 
-import "./interfaces/ISignalBuyContract.sol";
-import "./interfaces/IWETH.sol";
+import { ISignalBuyContract } from "./interfaces/ISignalBuyContract.sol";
+import { IWETH } from "./interfaces/IWETH.sol";
 import { Asset } from "./utils/Asset.sol";
-import "./utils/BaseLibEIP712.sol";
-import "./utils/LibConstant.sol";
-import "./utils/LibSignalBuyContractOrderStorage.sol";
-import "./utils/Ownable.sol";
+import { BaseLibEIP712 } from "./utils/BaseLibEIP712.sol";
+import { LibConstant } from "./utils/LibConstant.sol";
+import { LibSignalBuyContractOrderStorage } from "./utils/LibSignalBuyContractOrderStorage.sol";
+import { Ownable } from "./utils/Ownable.sol";
 import { Order, getOrderStructHash, Fill, getFillStructHash, AllowFill, getAllowFillStructHash } from "./utils/SignalBuyContractLibEIP712.sol";
-import "./utils/SignatureValidator.sol";
+import { SignatureValidator } from "./utils/SignatureValidator.sol";
 
 /// @title SignalBuy Contract
 /// @notice Order can be filled as long as the provided dealerToken/userToken ratio is better than or equal to user's specfied dealerToken/userToken ratio.
--- a/test/forkMainnet/SignalBuyContract.t.sol
+++ b/test/forkMainnet/SignalBuyContract.t.sol
@@ -480,14 +480,23 @@ contract SignalBuyContractTest is BalanceUtil {
         // Fill with default allow fill
         signalBuyContract.fillSignalBuy(DEFAULT_ORDER, DEFAULT_ORDER_MAKER_SIG, DEFAULT_TRADER_PARAMS, DEFAULT_CRD_PARAMS);
 
-        Fill memory fill = DEFAULT_FILL;
-        fill.dealerSalt = uint256(8001);
+        vm.expectRevert("SignalBuyContract: AllowFill seen before");
+        signalBuyContract.fillSignalBuy(DEFAULT_ORDER, DEFAULT_ORDER_MAKER_SIG, DEFAULT_TRADER_PARAMS, DEFAULT_CRD_PARAMS);
+    }
 
-        ISignalBuyContract.TraderParams memory traderParams = DEFAULT_TRADER_PARAMS;
-        traderParams.dealerSig = _signFill(dealerPrivateKey, fill, SignatureValidator.SignatureType.EIP712);
+    function testCannotFillByTraderWithReplayedFill() public {
+        // Fill with default allow fill
+        signalBuyContract.fillSignalBuy(DEFAULT_ORDER, DEFAULT_ORDER_MAKER_SIG, DEFAULT_TRADER_PARAMS, DEFAULT_CRD_PARAMS);
+
+        AllowFill memory allowFill = DEFAULT_ALLOW_FILL;
+        allowFill.salt = allowFill.salt + 1;
+
+        ISignalBuyContract.CoordinatorParams memory crdParams = DEFAULT_CRD_PARAMS;
+        crdParams.salt = allowFill.salt;
+        crdParams.sig = _signAllowFill(coordinatorPrivateKey, allowFill, SignatureValidator.SignatureType.EIP712);
 
         vm.expectRevert("SignalBuyContract: Fill seen before");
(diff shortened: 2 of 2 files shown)


charlesjhongc:
Overall no particular issues to me. I think the app can test on it first to see whether it fits the need or not.

contracts/PionexContract.sol
```diff
@@ -0,0 +1,632 @@
...
+                takerTokenAmount: _params.takerTokenAmount,
+                takerSalt: _params.salt,
+                expiry: _params.expiry
+            });
+            _validateTraderFill(fill, _params.takerSig);
+        }
+
+        (uint256 makerTokenAmount, uint256 takerTokenAmount, uint256 remainingAmount) = _quoteOrderFromMakerToken(_order, orderHash, _params.makerTokenAmount);
+        // Adjust takerTokenAmount according to the provided takerToken/makerToken ratio
+        takerTokenAmount = makerTokenAmount.mul(_params.takerTokenAmount).div(_params.makerTokenAmount);
```
pilagod, line 160: `takerTokenAmount` would be derived by exchange ratio, maybe we can skip `takerTokenAmount` calculation in `_quoteOrderFromMakerToken` to save some gas?

contracts/PionexContract.sol
```diff
@@ -0,0 +1,632 @@
...
+        view
+        returns (
+            uint256,
+            uint256,
+            uint256
+        )
+    {
+        uint256 makerTokenFilledAmount = LibOrderStorage.getStorage().orderHashToMakerTokenFilledAmount[_orderHash];
+
+        require(makerTokenFilledAmount < _order.makerTokenAmount, "LimitOrder: Order is filled");
```
pilagod, line 502: I'm pondering if an order could be filled twice, since by trader checks maker token filled, but by protocol checks taker token filled. 

In other words, an order could be first fully filled by trader, and it would only affect `orderHashToMakerTokenFilledAmount` record. And then the order could be filled again by protocol, since `orderHashToTakerTokenFilledAmount` is still `0` now, and it could pass the quote check phase.

Not sure if my understanding is correct 🤔

test/forkMainnet/PionexContract.t.sol
```diff
@@ -0,0 +1,1648 @@
...
+        vm.expectRevert("LimitOrder: Order is cancelled already");
+        vm.prank(user, user); // Only EOA
+        userProxy.toLimitOrder(payload);
+    }
+
+    /*********************************
+     *        Test: signing          *
+     *********************************/
+
+    // function testLimitOrderEIP712Sig() public {
```
pilagod, line 1461: nit: Could remove if not needed.

contracts/PionexContract.sol
NIC619: Fixed in 7da8bba

contracts/PionexContract.sol
NIC619: Fixed in f8b2d26

test/forkMainnet/PionexContract.t.sol
NIC619: Fixed in dfbec2a

charlesjhongc (commented):
I think it's ready for them to test. Perhaps can consider remove coordinator first so it's easier to onboard them.

contracts/PionexContract.sol
```diff
@@ -0,0 +1,366 @@
...
+        // memory cache
+        ISpender _spender = spender;
+        address _feeCollector = feeCollector;
+
+        // Calculate user fee (user receives pionex token so fee is charged in pionex token)
+        // 1. Fee for Tokenlon
+        uint256 tokenlonFee = _mulFactor(_settlement.pionexTokenAmount, tokenlonFeeFactor);
+        // 2. Fee for Pionex, including gas fee and strategy fee
+        uint256 pionexFee = _mulFactor(_settlement.pionexTokenAmount, _settlement.gasFeeFactor + _settlement.pionexStrategyFeeFactor);
+        uint256 pionexTokenForMaker = _settlement.pionexTokenAmount.sub(tokenlonFee).sub(pionexFee);
```
pilagod, line 229: Q: Would it possible that token after deducting fees makes the final exchange ratio less than `_order.pionexTokenAmount/_order.userTokenAmount`?

contracts/PionexContract.sol
```diff
@@ -0,0 +1,366 @@
...
+        // -> _params.pionexTokenAmount/_params.userTokenAmount >= _order.pionexTokenAmount/_order.userTokenAmount
+        require(
+            _params.pionexTokenAmount.mul(_order.userTokenAmount) >= _order.minPionexTokenAmount.mul(_params.userTokenAmount),
+            "PionexContract: pionex/user token ratio not good enough"
+        );
+        // Check gas fee factor and pionex strategy fee factor do not exceed limit
+        require(
+            (_params.gasFeeFactor <= LibConstant.BPS_MAX) &&
+                (_params.pionexStrategyFeeFactor <= LibConstant.BPS_MAX) &&
+                (_params.gasFeeFactor + _params.pionexStrategyFeeFactor <= LibConstant.BPS_MAX - tokenlonFeeFactor),
```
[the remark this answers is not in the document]
NIC619: This is to prevent the sum from overflow so I added the first two checks.

contracts/PionexContract.sol
```diff
@@ -0,0 +1,366 @@
...
+        uint256 pionexTokenAmount = userTokenAmount.mul(_params.pionexTokenAmount).div(_params.userTokenAmount);
+
+        uint256 userTokenOut = _settleForTrader(
+            TraderSettlement({
+                orderHash: orderHash,
+                allowFillHash: allowFillHash,
+                trader: _params.pionex,
+                recipient: _params.recipient,
+                user: _order.user,
+                pionex: _order.pionex,
```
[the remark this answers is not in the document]
NIC619: I rename the `taker` to `pionex` so it might seem weird.🤣
- `_order.pionex` is `_order.taker`, included in the payload signed by maker
- `_params.pionex` is `_params.taker`, included in the payload signed by taker (Pionex) itself
- `_settlement.pionex` is `_settlement.taker`, I think in `LimitOrder` taker is also trader so `_settlement.taker` was not used in `_settleForTrader`: https://github.com/consenlabs/tokenlon-contracts/blob/ed4b74057268b4b8aeaf9fceafae5939ed01e413/contracts/LimitOrder.sol#L258

contracts/PionexContract.sol
NIC619: Yeah it is possible. We haven't decide on whether we should enforce that the eventual ratio is at least as good as the one user specified in the order. Since the gas fee will inevitably reduce the ratio and if Pionex provide the same ratio as the one specified by user, this would result in the eventual ratio worse than the user specified.

contracts/PionexContract.sol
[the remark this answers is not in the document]
pilagod: I see. And should we remove `pionex` in the `TraderSettlement`, since it is not used by other place?

pilagod (dismissed):

contracts/PionexContract.sol
```diff
@@ -0,0 +1,367 @@
...
+    ) external override onlyUserProxy nonReentrant returns (uint256, uint256) {
+        bytes32 orderHash = getEIP712Hash(PionexContractLibEIP712._getOrderStructHash(_order));
+
+        _validateOrder(_order, orderHash, _orderUserSig);
+        bytes32 allowFillHash = _validateFillPermission(orderHash, _params.pionexTokenAmount, _params.pionex, _crdParams);
+        _validateOrderTaker(_order, _params.pionex);
+
+        // Check provided pionexToken/userToken ratio is better than or equal to user's specfied pionexToken/userToken ratio
+        // -> _params.pionexTokenAmount/_params.userTokenAmount >= _order.pionexTokenAmount/_order.userTokenAmount
+        require(
```
charlesjhongc, line 111: Since the `minPionexTokenAmount` should be checked after fee deducted, may be this require can be removed since it'd be reverted if `pionexTokenForUser < _settlement.minPionexTokenAmount` even the ratio here is better?

NIC619: Good call. Fixed in ed48d9b

contracts/SignalBuyContract.sol
```diff
@@ -0,0 +1,411 @@
...
+pragma abicoder v2;
+
+import "@openzeppelin/contracts/math/Math.sol";
+import "@openzeppelin/contracts/math/SafeMath.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+import "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+
+import "./interfaces/ISignalBuyContract.sol";
+import "./interfaces/IPermanentStorage.sol";
```
charlesjhongc, line 12: Can be removed.

contracts/interfaces/ISignalBuyContract.sol
```diff
@@ -0,0 +1,91 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.7.0;
+pragma abicoder v2;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
charlesjhongc, line 5: Unused?

charlesjhongc (approved):

changwu-tw (dismissed):

changwu-tw (dismissed):

Jpk313183 (approved):
Approving the merging

PR #198, opened 2023-05-19
Adjust deployed test

charlesjhongc:
- merge `test/utils/Tokens` into `test/tuils/Addresses`
- remove envrc and read json file instead
- make some ERC20 address as private to avoid misreading


test/utils/Addresses.sol
```diff
@@ -1,23 +1,45 @@
...
+    IERC20 public ankreth;
+    IERC20[] public tokens;
+
     string private file = readAddresses(vm);
 
-    address WETH_ADDRESS = abi.decode(vm.parseJson(file, "$.WETH_ADDRESS"), (address));
-    address USDT_ADDRESS = abi.decode(vm.parseJson(file, "$.USDT_ADDRESS"), (address));
-    address USDC_ADDRESS = abi.decode(vm.parseJson(file, "$.USDC_ADDRESS"), (address));
+    // Since token may be newly deployed in local testnet case, these imported addresses may not be the actual ones.
+    address private WETH_ADDRESS = abi.decode(vm.parseJson(file, "$.WETH_ADDRESS"), (address));
```
charlesjhongc, line 25: I know that it mixes `private` and `public` among these addresses but currently I don't have a better solution.

test/utils/BalanceUtil.sol
```diff
@@ -17,7 +18,11 @@ contract BalanceUtil is Addresses {
     ) internal {
         uint256 decimals = uint256(ERC20(tokenAddr).decimals());
         // Skip setting WETH's totalSupply because WETH does not store total supply in storage
-        bool updateTotalSupply = tokenAddr == WETH_ADDRESS ? false : true;
+        // Only update WETH's totalSupply in local env
+        bool updateTotalSupply = false;
+        if (tokenAddr == address(weth) && getChainId() == 31337) {
+            updateTotalSupply = true;
+        }
```
pilagod, line 25:
```suggestion
        bool updateTotalSupply = getChainId() == 31337 && tokenAddr == address(weth);
```

test/utils/Addresses.sol
pilagod: Do you mean that `TOKEN_ADDRESS` is private, and `token` is public? What goal would you want to achieve?

pilagod: Do you mean that `TOKEN_ADDRESS` is private, and `token` is public? What goal would you want to achieve?

charlesjhongc: Yeah, the goal of doing this is to unify the way of reading token address. For example a test can read WETH address by `WETH_ADDRESS` or `address(weth). But for local test, tokens are new deployed mock tokens so `WETH_ADDRESS` is not the read address of mock WETH.

charlesjhongc: However, this only applies to some of tokens so there're private and public `TOKEN_ADDRESS`.

NIC619 (dismissed):

pilagod (approved):

NIC619 (approved):

PR #200, opened 2023-05-22
Integrate TokenCollector with AllowanceTarget

charlesjhongc:


test/forkMainnet/GenericSwap.t.sol
```diff
@@ -59,9 +61,15 @@ contract GenericSwapTest is Test, Tokens, BalanceUtil {
     GenericSwap genericSwap;
     GenericSwapData gsData;
     MockStrategy mockStrategy;
+    AllowanceTarget allowanceTarget;
 
     function setUp() public {
-        genericSwap = new GenericSwap(UNISWAP_PERMIT2_ADDRESS);
+        // deploy allowance target
+        address[] memory trusted = new address[](1);
+        trusted[0] = computeContractAddress(address(this), uint8(vm.getNonce(address(this)) + 1));
```
NIC619, line 69: Maybe add a comment to explain why we are doing this?

NIC619 (approved):

pilagod (approved):

test/forkMainnet/GenericSwap.t.sol
charlesjhongc: It's because the white list is immutable so have to pre-compute the contract address. Added in a07234b.

pilagod: I think it is more relevant to the bi-directional dependency between allowance target and the strategy contract (they need each other in order to deploy). Although imo it would be better to decouple this kind of dependency, I have no good idea to refine it at this moment 😅

pilagod (dismissed):

test/forkMainnet/GenericSwap.t.sol
NIC619: Sorry my comment didn't provide enough detail. I mainly want to make this behavior and its assumption explicit so others can be aware of and not breaking it when they copy or modify from it in the future.

- What is the address we are computing here? `GeneraicSwap`
- What is the assumption? `GenericSwap` must be deployed right next to `AllowanceTarget`

Suggested change:
```solidity
// pre-compute GenericSwap address since the whitelist of allowance target is immutable
// NOTE: this assumes GenericSwap is deployed right next to Allowance Target
address GSAddress = computeContractAddress(address(this), uint8(vm.getNonce(address(this)) + 1));
trusted[0] = GSAddress;
```

Also others can more quickly understand what this step does and more easily extend it to multiple `trusted` addresses.

charlesjhongc: No problem. Adjusted in 9e4edfd.

NIC619 (approved):

pilagod (approved):

PR #201, opened 2023-05-23
check EIP712 chain ID

charlesjhongc:


pilagod (commented):
If the fork does happen, it would fallback to re-compute separator every time when hashing, which might increase lots of gas cost. (under my impression sometimes we would call hash multiple times in one tx)

Although fork seldom happens, do you think it is worth to upgrade immutable to storage, and update them when needed according to Decurity's suggestion?

`If a change in chainID is detected, it is possible to cache and renew the domain separator to 
mitigate  this  risk`

charlesjhongc:
@pilagod Good point. I think it depends on our policy of forked chain. If we will operate on the forked chain, then updating domain separator is making sense. Currently I think we don't and the main reason is avoid signature being replayed in order to protect users. Therefore I think the immutable is more suitable since it's more gas efficient for mainnet users.

contracts/utils/BaseLibEIP712.sol
```diff
@@ -10,16 +10,13 @@ abstract contract BaseLibEIP712 {
...
-        keccak256(
-            abi.encode(
-                keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"),
-                keccak256(bytes(EIP712_DOMAIN_NAME)),
-                keccak256(bytes(EIP712_DOMAIN_VERSION)),
-                getChainID(),
-                address(this)
-            )
-        );
+    bytes32 public immutable EIP712_DOMAIN_SEPARATOR;
```
NIC619, line 13: `EIP712_DOMAIN_SEPARATOR` probably should use the same fallback mechanism in `_getDomainSeparator`? In case dApps just read the variable `EIP712_DOMAIN_SEPARATOR`.

```solidity
bytes32 private immutable originalEIP712DomainSeparator;

constructor() {
        originalEIP712DomainSeparator = _buildDomainSeparator();
        originalChainId = getChainID();
}

function _getDomainSeparator() private view returns (bytes32) {
        if (getChainID() == originalChainId) {
            return originalEIP712DomainSeparator;
        } else {
            return _buildDomainSeparator();
        }
}

function EIP712_DOMAIN_SEPARATOR() external view returns (bytes32) {
        return _getDomainSeparator();
}
```

charlesjhongc: Yeah it should although in most cases dapp should use `getEIP712Hash()` directly. Fixed in 246d8d1.

NIC619 (approved):

changwu-tw (approved):

pilagod (approved):

PR #204, opened 2023-05-23
RFQv2 optimize uint comparison

charlesjhongc:


changwu-tw (approved):

contracts/RFQv2.sol
```diff
@@ -72,7 +72,7 @@ contract RFQv2 is IRFQv2, StrategyBase, TokenCollector, SignatureValidator, Base
         Offer memory _offer = _rfqOrder.offer;
         // check the offer deadline and fee factor
         require(_offer.expiry >= block.timestamp, "offer expired");
```
pilagod, line 74: nit: Should we also change this line?

pilagod (approved):

contracts/RFQv2.sol
charlesjhongc: Yeah I think so. Fixed in 82bca93.

pilagod (approved):

NIC619 (approved):

PR #209, opened 2023-05-25
Apply latest audit fixes to v6

charlesjhongc:


contracts/libraries/Asset.sol
```diff
@@ -27,9 +26,14 @@ library Asset {
...
+        if (to == address(this) || amount == 0) {
+            return;
+        }
         if (isETH(asset)) {
-            // FIXME replace with fixed gas to solve reentrancy issue
-            Address.sendValue(to, amount);
+            // @dev forward all available gas and may cause reentrancy
+            require(address(this).balance >= amount, "insufficient balance");
+            (bool success, ) = to.call{ value: amount }("");
+            require(success, "unable to send ETH");
```
pilagod, line 36: Q: So should we limit the gas? If not, why would us switch to this implementation

charlesjhongc: The idea of limiting gas is to prevent reentrancy, but how much gas should be enough is a problem since gas usage of EVM instructions may be [altered from time to time](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Address.sol#L48-L69). Also, it's not the only way to prevent reentrancy. Others like `ReentrancyGuard` by OZ or carefully handle the order of state changes and external calls (or ETH transfer). In fact this design is used by v5 RFQv2 already so basically this PR is just applying v5 changes to v6.

pilagod: Thanks for the explanation 🫡 I might miss this change in the v5 side, my bad 😢

pilagod (approved):

contracts/libraries/Asset.sol
charlesjhongc: All good man. It's good to review changes even they're the same but with different merge target.

NIC619 (approved):

PR #212, opened 2023-05-29
Re Implement RFQ partial fill

charlesjhongc:


contracts/libraries/RFQTx.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { RFQOffer, getRFQOfferHash, RFQ_OFFER_TYPESTRING } from "./RFQOffer.sol";
+
+string constant RFQ_TX_TYPESTRING = string(
+    abi.encodePacked("RFQTx(RFQOffer rfqOffer,bytes32 offerHash,address recipient,uint256 takerRequestAmount,uint256 feeFactor)", RFQ_OFFER_TYPESTRING)
```
NIC619, line 7: Do we need both `rfqOffer` and `offerHash`  here?

contracts/interfaces/IGenericSwap.sol
```diff
@@ -5,7 +5,6 @@ import { GenericSwapData } from "../libraries/GenericSwapData.sol";
 
 interface IGenericSwap {
     error AlreadyFilled();
-    error InvalidTaker();
```
NIC619, line 8: We are not checking this anymore?

contracts/GenericSwap.sol
```diff
@@ -53,41 +52,27 @@ contract GenericSwap is IGenericSwap, TokenCollector, EIP712 {
...
-            swapData.offer.takerTokenAmount,
-            swapData.offer.makerToken,
-            returnAmount
-        );
+        emit Swap(swapHash, swapData.maker, taker, swapData.recipient, swapData.takerToken, swapData.takerTokenAmount, swapData.makerToken, returnAmount);
     }
 
     function _executeSwap(
         GenericSwapData memory _swapData,
         address _authorizedUser,
```
NIC619, line 60: Maybe rename to taker?

pilagod (dismissed):

test/forkMainnet/RFQ.t.sol
```diff
@@ -254,173 +273,257 @@ contract RFQTest is Test, Tokens, BalanceUtil {
...
+    function testFillWithContract() public {
+        RFQOffer memory rfqOffer = defaultRFQOffer;
+        rfqOffer.flags |= FLG_ALLOW_CONTRACT_SENDER;
+        bytes memory makerSig = _signRFQOffer(makerSignerPrivateKey, rfqOffer);
+
+        RFQTx memory rfqTx = defaultRFQTx;
+        rfqTx.rfqOffer = rfqOffer;
+
+        vm.prank(rfqOffer.taker, makeAddr("anyAddr"));
+        rfq.fillRFQ(rfqTx, makerSig, defaultPermit, defaultPermit);
```
NIC619, line 319: nit:
```suggestion
        rfq.fillRFQ(rfqTx, makerSig, defaultPermit, defaultPermit, takerSignature);
```

contracts/libraries/RFQTx.sol
charlesjhongc: The `offerHash` should be removed. f6f45a6

contracts/interfaces/IGenericSwap.sol
charlesjhongc: Since GS order doesn't have maker sig, checking taker is not making too much sense here actually.

contracts/GenericSwap.sol
charlesjhongc: I think the `authorized` word is required so it's more clear for dev that this address has already been verified.

test/forkMainnet/RFQ.t.sol
charlesjhongc: `Fill with contract` is about the `msg.sender` should be a contract. It's not directly relevant that with or without taker signature.

contracts/GenericSwap.sol
```diff
@@ -27,12 +26,12 @@ contract GenericSwap is IGenericSwap, TokenCollector, EIP712 {
         returnAmount = _executeSwap(swapData, msg.sender, takerTokenPermit);
         emit Swap(
             getGSDataHash(swapData),
-            swapData.offer.maker,
-            swapData.offer.taker,
+            swapData.maker,
+            msg.sender,
```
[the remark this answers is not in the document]
charlesjhongc: Added in add061d

contracts/GenericSwap.sol
NIC619: I thought this address is not verified anymore?

test/forkMainnet/RFQ.t.sol
NIC619: Yeah what I mean is that the examples in our tests served as a reference for others to see e.g., how to interact with our contract and so it would be helpful to mimic the realistic behavior. In this scenario `msg.sender` should be a contract and the function it should be calling should the one with taker signature attached.

contracts/GenericSwap.sol
charlesjhongc: Either the address == `msg.sender`, or it's authorized by the taker signature. One cannot trade on other's behalf without proper authorization.

test/forkMainnet/RFQ.t.sol
charlesjhongc: What is the scenario of calling with a contract but trade on an EOA account's behalf? One real example of this test case is a multi-sig wallet, so the taker would be the multi-sig contract.

NIC619: The test case is `testFillWithContract` so it would be a little weird that we prank an EOA as `msg.sender`. The test passes as long we prank different `msg.sender` and `tx.origin` but that's what I meant about mimic realistic behavior - in reality a contract will be the `msg.sender` and in this case it would be calling the `fillRFQ` function with taker signature attached, not the other `fillRFQ` function.

charlesjhongc: Oh I see, my bad.

charlesjhongc: Adjusted in a4f0e46.

NIC619 (approved):

pilagod (approved):

PR #215, opened 2023-05-30
Implement v6 limit order

charlesjhongc:
- Separate `coordinator` from `LimitOrder` contract, implement it using a special taker contact instead
- Add `fillLimitOrderFullOrKill` function to allow taker choose different fill strategy
- Allow fill with better price than the order specified
- Add `GroupFill` feature
- `FeeFactor` now is per order, not global


charlesjhongc:
The light weight version #221

PR #221, opened 2023-06-02
Implement v6 limit order base

charlesjhongc:
- Add `fillLimitOrderFullOrKill` function to allow taker choose different fill strategy
- Allow fill with better price than the order specified
- `FeeFactor` now is per order, not global


contracts/interfaces/ILimitOrderSwap.sol
```diff
@@ -0,0 +1,66 @@
...
+    error CanceledOrder();
+    error FilledOrder();
+    error ZeroAddress();
+    error ZeroTokenAmount();
+    error NotEnoughForFill();
+    error InvalidMsgValue();
+    error InvalidSignature();
+    error InvalidTaker();
+    error InvalidTakingAmount();
+    error InvalidParams();
```
NIC619, line 19: This does not seem to be used.

test/forkMainnet/LimitOrderSwap/Fill.t.sol
```diff
@@ -0,0 +1,527 @@
...
+    address[] defaultAMMPath = [DAI_ADDRESS, USDT_ADDRESS];
+
+    function testFullyFillLimitOrder() public {
+        Snapshot memory takerTakerToken = BalanceSnapshot.take({ owner: taker, token: defaultOrder.takerToken });
+        Snapshot memory takerMakerToken = BalanceSnapshot.take({ owner: taker, token: defaultOrder.makerToken });
+        Snapshot memory makerTakerToken = BalanceSnapshot.take({ owner: defaultOrder.maker, token: defaultOrder.takerToken });
+        Snapshot memory makerMakerToken = BalanceSnapshot.take({ owner: defaultOrder.maker, token: defaultOrder.makerToken });
+        Snapshot memory recTakerToken = BalanceSnapshot.take({ owner: recipient, token: defaultOrder.takerToken });
+        Snapshot memory recMakerToken = BalanceSnapshot.take({ owner: recipient, token: defaultOrder.makerToken });
+        Snapshot memory feeCollectorBal = BalanceSnapshot.take({ owner: feeCollector, token: defaultOrder.makerToken });
```
NIC619, line 23: nit: `feeCollectorMakerToken` or `fcMakerToken`?

test/forkMainnet/LimitOrderSwap/Fill.t.sol
```diff
@@ -0,0 +1,527 @@
...
+        takerTakerToken.assertChange(-int256(defaultOrder.takerTokenAmount));
+        takerMakerToken.assertChange(int256(0));
+        makerTakerToken.assertChange(int256(defaultOrder.takerTokenAmount));
+        makerMakerToken.assertChange(-int256(defaultOrder.makerTokenAmount));
+        recTakerToken.assertChange(int256(0));
+        recMakerToken.assertChange(int256(defaultOrder.makerTokenAmount - fee));
+        feeCollectorBal.assertChange(int256(fee));
+    }
+
+    function testFullyFillLimitOrderUsingAMM() public {
```
NIC619, line 52: Add a test for strategy failing to deliver enough taker token amount?

contracts/LimitOrderSwap.sol
```diff
@@ -0,0 +1,207 @@
...
+            takerParams.makerTokenAmount,
+            fullOrKill
+        );
+
+        // maker -> taker
+        _collect(order.makerToken, order.maker, address(this), makerSpendingAmount, order.makerTokenPermit);
+        uint256 fee = (makerSpendingAmount * order.feeFactor) / Constant.BPS_MAX;
+        order.makerToken.transferTo(payable(takerParams.recipient), makerSpendingAmount - fee);
+        // collect fee if present
+        order.makerToken.transferTo(feeCollector, fee);
```
pilagod, line 104: nit: Not sure if checking fee > 0 before transferring could save some gas for no fee case? But if we always have fee, then this might not worth to add. 😆

contracts/LimitOrderSwap.sol
```diff
@@ -0,0 +1,207 @@
...
+            // the requested amount is larget than fillable amount
+            if (_fullOrKill) {
+                revert NotEnoughForFill();
+            } else {
+                // take the rest of this order
+                makerSpendingAmount = orderAvailableAmount;
+
+                // re-calculate the amount of taker willing to spend for this trade by the requested ratio
+                _takerTokenAmount = ((_takerTokenAmount * makerSpendingAmount) / _makerTokenAmount);
+            }
```
pilagod, line 171: nit: This else block could be unwrap to reduce the indentation.

test/mocks/MockLimitOrderTaker.sol
```diff
@@ -0,0 +1,53 @@
...
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import { Ownable } from "contracts/abstracts/Ownable.sol";
+import { IStrategy } from "contracts/interfaces/IStrategy.sol";
+import { IUniswapRouterV2 } from "contracts/interfaces/IUniswapRouterV2.sol";
+import { MockERC1271Wallet } from "./MockERC1271Wallet.sol";
+
+contract MockLimitOrderTaker is IStrategy, MockERC1271Wallet {
```
pilagod, line 12: nit: I think in the future we could use a simpler mock contract instead of integrating real protocol, since we already decided to mock it 😆

contracts/interfaces/ILimitOrderSwap.sol
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /// @notice Fill an order
+    function fillLimitOrder(
+        LimitOrder calldata order,
+        bytes calldata makerSignature,
+        TakerParams calldata takerParams
+    ) external payable;
+
+    /// @notice Fill an order
```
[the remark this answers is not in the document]
charlesjhongc: Yeah, after the implementation is confirmed.

contracts/LimitOrderSwap.sol
```diff
@@ -0,0 +1,207 @@
...
+        } else {
+            // the requested amount can be statisfied
+            makerSpendingAmount = _makerTokenAmount;
+        }
+        uint256 minTakerTokenAmount = ((makerSpendingAmount * _order.takerTokenAmount) / _order.makerTokenAmount);
+        // check if taker provide enough amount for this fill (better price is allowed)
+        if (_takerTokenAmount < minTakerTokenAmount) revert InvalidTakingAmount();
+        takerSpendingAmount = _takerTokenAmount;
+
+        if (takerSpendingAmount == 0 && makerSpendingAmount == 0) revert ZeroTokenAmount();
```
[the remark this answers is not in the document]
charlesjhongc: My original thought is `AND` since it's possible for some order that one of these amount is zero. It may make no sense but would be no harm to have such flexibility. But if both are all zero than there's no need to proceed.

contracts/LimitOrderSwap.sol
```diff
@@ -0,0 +1,207 @@
...
+
+        // validate maker signature only once per order
+        if (orderFilledAmount == 0) {
+            if (!SignatureValidator.isValidSignature(_order.maker, getEIP712Hash(orderHash), _makerSignature)) revert InvalidSignature();
+        }
+
+        // get the quote of the fill
+        uint256 orderAvailableAmount = _order.makerTokenAmount - orderFilledAmount;
+        if (_makerTokenAmount > orderAvailableAmount) {
+            // the requested amount is larget than fillable amount
```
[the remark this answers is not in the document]
charlesjhongc: Fixed in 9f12b2b

contracts/interfaces/ILimitOrderSwap.sol
```diff
@@ -0,0 +1,66 @@
...
+    error InvalidParams();
+    error NotOrderMaker();
+
+    /// @notice Emitted when fee collector address is updated
+    /// @param newFeeCollector The address of the new fee collector
+    event SetFeeCollector(address newFeeCollector);
+
+    /// @notice Emitted when an order is filled
+    event LimitOrderFilled(
+        bytes32 indexed offerHash,
```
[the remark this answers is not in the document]
charlesjhongc: Fixed in 9f12b2b

contracts/interfaces/ILimitOrderSwap.sol
charlesjhongc: Fixed in 9f12b2b

test/forkMainnet/LimitOrderSwap/Fill.t.sol
charlesjhongc: Fixed in 9f12b2b

test/forkMainnet/LimitOrderSwap/Fill.t.sol
charlesjhongc: Actually it depends on the logic of strategy contract. A simple case could be that taker doesn't have any taker tokens and only relies on strategy contract. Then if strategy contract not returning enough to taker then the collecting taker token from taker will fail (the last `transferFrom`).

Added in 2b9cb7d.

test/mocks/MockLimitOrderTaker.sol
charlesjhongc: Yeah actually it's more like integration test to be honest. I also added the mock strategy contract (shared with GS) so we can separate these two cases as different type of test in the future.

contracts/LimitOrderSwap.sol
```diff
@@ -0,0 +1,207 @@
...
+contract LimitOrderSwap is ILimitOrderSwap, Ownable, TokenCollector, EIP712 {
+    using Asset for address;
+
+    IWETH public immutable weth;
+    address payable public feeCollector;
+
+    // how much maker token has been filled in an order
+    mapping(bytes32 => uint256) public orderHashToMakerTokenFilledAmount;
+    // whether an order is canceled or not
+    mapping(bytes32 => bool) public orderHashToCanceled;
```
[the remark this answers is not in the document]
charlesjhongc: Good idea. Refactored in af642a7.

NIC619 (approved):

pilagod (approved):

PR #222, opened 2023-06-04
Implement UniAgent contract

charlesjhongc:
After several tests of different versions, I found that it's not worthy to have the relayed design in here. From the gas efficiency perspective, the relayed one cost similar as GS does so no need to have another similar implementation. However, the self-submit way could perform much closer as user directly call the uniswap official router.

### DIFF
--- a/contracts/UniAgent.sol
+++ b/contracts/UniAgent.sol
@@ -80,7 +80,7 @@ contract UniAgent is Ownable, IUniAgent, TokenCollector, EIP712 {
         if (!inputToken.isETH()) {
             if (msg.value != 0) revert InvalidMsgValue();
 
-            if (routerType == RouterType.universal) {
+            if (routerType == RouterType.UniversalRouter) {
                 // deposit directly into router if it's universal router
                 _collect(inputToken, msg.sender, universalRouter, inputAmount, userPermit);
             } else {
@@ -99,11 +99,11 @@ contract UniAgent is Ownable, IUniAgent, TokenCollector, EIP712 {
     }
 
     function _getRouterAddress(RouterType routerType) private pure returns (address) {
-        if (routerType == RouterType.v2) {
+        if (routerType == RouterType.V2Router) {
             return v2Router;
-        } else if (routerType == RouterType.v3) {
+        } else if (routerType == RouterType.V3Router) {
             return v3Router;
-        } else if (routerType == RouterType.universal) {
+        } else if (routerType == RouterType.UniversalRouter) {
             return universalRouter;
         } else {
             revert UnknownRouterType();
--- a/contracts/UniAgent.sol
+++ b/contracts/UniAgent.sol
@@ -17,8 +17,6 @@ contract UniAgent is Ownable, IUniAgent, TokenCollector, EIP712 {
     using Asset for address;
     using SafeERC20 for IERC20;
 
-    uint256 private constant FEE_ON_INPUT_MASK = 1 << 255;
-    uint256 private constant FEE_AMOUNT_MASK = 2**255 - 1;
     address private constant v2Router = 0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D;
     address private constant v3Router = 0xE592427A0AEce92De3Edee1F18E0157C05861564;
     address payable private constant universalRouter = payable(0xEf1c6E67703c7BD7107eed8303Fbe6EC2554BF6B);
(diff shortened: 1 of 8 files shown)


contracts/UniAgent.sol
```diff
@@ -0,0 +1,112 @@
...
+        if (routerType == RouterType.v2) {
+            return v2Router;
+        } else if (routerType == RouterType.v3) {
+            return v3Router;
+        } else if (routerType == RouterType.universal) {
+            return universalRouter;
+        } else {
+            revert UnknownRouterType();
+        }
+    }
```
pilagod, line 111: I thought universal router already put v2 and v3 together, do we need additional logic to get specific router address?

contracts/interfaces/IUniAgent.sol
```diff
@@ -0,0 +1,26 @@
...
+    error ZeroAddress();
+    error InvalidMsgValue();
+    error UnknownRouterType();
+
+    event Swap(address indexed user, address indexed router, address indexed inputToken, uint256 inputAmount);
+
+    enum RouterType {
+        v2,
+        v3,
+        universal
```
pilagod, line 16: nit: I personally would prefer using pascal case for enum key.

contracts/UniAgent.sol
charlesjhongc: From functionality perspective, we don't need v2 & v3 router if we have universal router. But from gas efficiency perspective, it could save some gas in some case if calling v2 or v3 router directly. The major reason is that decoding input data is needed in UA and it supports multicall. From my rough observation, only txs composed by at least two steps will be dispatched to UA. In another words, for those tx only take one single step, v2 & v3 router is a better choice.

contracts/interfaces/IUniAgent.sol
charlesjhongc: Ok, should be easier to identify. Adjusted in f1d1078.

pilagod (dismissed):

contracts/UniAgent.sol
```diff
@@ -0,0 +1,112 @@
...
+    function _getRouterAddress(RouterType routerType) private pure returns (address) {
+        if (routerType == RouterType.V2Router) {
+            return v2Router;
+        } else if (routerType == RouterType.V3Router) {
+            return v3Router;
+        } else if (routerType == RouterType.UniversalRouter) {
+            return universalRouter;
+        } else {
+            revert UnknownRouterType();
+        }
```
NIC619, line 110: I think solidity would catch the invalid enum value and just the fail the transaction.

contracts/UniAgent.sol
```diff
@@ -0,0 +1,112 @@
...
+    receive() external payable {}
+
+    function setFeeCollector(address payable _newFeeCollector) external onlyOwner {
+        if (_newFeeCollector == address(0)) revert ZeroAddress();
+        feeCollector = _newFeeCollector;
+
+        emit SetFeeCollector(_newFeeCollector);
+    }
+
+    function approveTokens(address[] calldata tokens, address[] calldata spenders) external onlyOwner {
```
NIC619, line 53: This would require owner to approve all tokens that we allow users to trade?

charlesjhongc: Yes. Do you think that users should be able to approve tokens by themselves?

charlesjhongc: Also, I think the `spenders` can be remove can approve 3 router for each token.

charlesjhongc: How about adding a `approveAndSwap` function? So if a user wants to trade a new token pair, they can do it within one tx.

charlesjhongc: Or adding a `bool` flag

NIC619: My concern is actually about owner, i.e., us, having to approve again every time users request a new token for them to trade.

And my thought was actually to automatically approve the token amount to be swapped in V2/V3 cases:
```solidity
            if (routerType == RouterType.UniversalRouter) {
                ...
            } else {
                // v2 v3 use transferFrom
                _collect(inputToken, msg.sender, address(this), inputAmount, userPermit);
                inputToken.safeApprove(routerAddr, inputAmount);
            }
```
Does this approach work or perhaps I'm missing something?

NIC619: `approveAndSwap` is probably better than a `bool` flag as user or his wallet does not have to simulate or do a pre-check before every swap. Especially when the `allowance not enough` error is not very informative for user to debug in my opinion.

charlesjhongc: Added the `approveAndSwap` in a5e1624.

NIC619 (approved):

pilagod (approved):

PR #226, opened 2023-06-07
Refactor token collector

charlesjhongc:
~To shorten the data for token collector, add `length == 1` condition for basic token source type.~
Use bytes concat for permit data wrapping. First it's easier to detect the type of source without abi decoding, and second it saves some length of data.

### DIFF
--- a/contracts/abstracts/TokenCollector.sol
+++ b/contracts/abstracts/TokenCollector.sol
@@ -11,9 +11,12 @@ import { IAllowanceTarget } from "../interfaces/IAllowanceTarget.sol";
 abstract contract TokenCollector {
     using SafeERC20 for IERC20;
 
+    error UnknownTokenSource();
+
     enum Source {
-        Token,
         TokenlonAllowanceTarget,
+        Token,
+        TokenPermit,
         Permit2AllowanceTransfer,
         Permit2SignatureTransfer
     }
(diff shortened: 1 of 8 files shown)


charlesjhongc:
Call for feedback, will fix the rest once confirmed.

charlesjhongc:
Pushed a new commit but PR was not updated. Reopen a new one.

test/forkMainnet/TokenCollector.t.sol
```diff
@@ -120,7 +120,11 @@ contract TestTokenCollector is Addresses {
         bytes32 r,
         bytes32 s
     ) private pure returns (bytes memory) {
-        return abi.encode(TokenCollector.Source.Token, abi.encode(permit.owner, permit.spender, permit.amount, permit.deadline, v, r, s));
+        return
+            bytes.concat(
+                abi.encodePacked(TokenCollector.Source.TokenPermit),
+                abi.encode(permit.owner, permit.spender, permit.amount, permit.deadline, v, r, s)
```
pilagod, line 126: Q: Cannot just `abi.encodePacked(TokenPermit, abi.encode( ... ))`?

charlesjhongc: It works as well. Adjusted in 4b13f26.

contracts/abstracts/TokenCollector.sol
```diff
@@ -31,87 +35,50 @@ abstract contract TokenCollector {
...
-        if (data.length > 0) {
-            (uint48 nonce, uint48 deadline, bytes memory permitSig) = abi.decode(data, (uint48, uint48, bytes));
-            IUniswapPermit2.PermitSingle memory permit = IUniswapPermit2.PermitSingle({
-                details: IUniswapPermit2.PermitDetails({ token: token, amount: uint160(amount), expiration: deadline, nonce: nonce }),
-                spender: address(this),
-                sigDeadline: uint256(deadline)
+            return IERC20(token).safeTransferFrom(from, to, amount);
+        } else if (src == Source.Permit2AllowanceTransfer) {
+            if (amount > uint256(type(uint160).max)) revert Permit2AmountTooLarge();
+            bytes memory permit2Data = data[1:];
```
NIC619, line 57: Can it also be `calldata`?

test/forkMainnet/TokenCollector.t.sol
```diff
@@ -120,7 +120,10 @@ contract TestTokenCollector is Addresses {
         bytes32 r,
         bytes32 s
     ) private pure returns (bytes memory) {
-        return abi.encode(TokenCollector.Source.Token, abi.encode(permit.owner, permit.spender, permit.amount, permit.deadline, v, r, s));
+        return
+            bytes.concat(
```
NIC619, line 124: Is it concatenating `abi.encodePacked(TokenCollector.Source.TokenPermit)` and `abi.encode(permit.owner, ...)`?

test/forkMainnet/TokenCollector.t.sol
pilagod: I think we could get rid of `bytes.concat` also?

contracts/abstracts/TokenCollector.sol
charlesjhongc: It will be copied to memory when doing abi decode so I think no huge different here. Also it will cause stack too deep.

test/forkMainnet/TokenCollector.t.sol
charlesjhongc: Yes, actually it can be replaced by a single `abi.encodePacked`.

test/forkMainnet/TokenCollector.t.sol
charlesjhongc: Yeah, fixed in 0f31586

test/forkMainnet/TokenCollector.t.sol
charlesjhongc: 0f31586

pilagod (approved):

NIC619 (approved):

PR #227, opened 2023-06-07
Refactor UniAgent and add NatSpec

charlesjhongc:
Token will be transferred to `UniversalRouter` directly so no need to approve.

### DIFF
--- a/contracts/interfaces/IUniAgent.sol
+++ b/contracts/interfaces/IUniAgent.sol
@@ -4,7 +4,6 @@ pragma solidity >=0.8.0;
 /// @title IUniAgent Interface
 /// @author imToken Labs
 interface IUniAgent {
-    error ZeroAddress();
     error InvalidMsgValue();
     error UnknownRouterType();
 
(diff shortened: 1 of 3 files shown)


contracts/interfaces/IUniAgent.sol
```diff
@@ -4,18 +4,29 @@ pragma solidity >=0.8.0;
 /// @title IUniAgent Interface
 /// @author imToken Labs
 interface IUniAgent {
-    error ZeroAddress();
     error InvalidMsgValue();
     error UnknownRouterType();
```
NIC619, line 8: I think we don't need this error because solidity automatically checks the validity of enum input for us (it would just raise EVM error. So it would never reach the code path here in `_getRouterAddress`:
```solidity
    function _getRouterAddress(...) ... {
        ...
        } else {
            revert UnknownRouterType();
        }
    }
```

NIC619 (approved):

contracts/interfaces/IUniAgent.sol
charlesjhongc: Removed the error but in order to avoid compiler warning (uncovered code path), revert without msg as default.

charlesjhongc: 5ae24e4

NIC619 (approved):

pilagod (approved):

PR #229, opened 2023-06-09
remove unwrapAmount and check WETH amount only

charlesjhongc:


contracts/LimitOrderSwap.sol
```diff
@@ -92,6 +92,10 @@ contract LimitOrderSwap is ILimitOrderSwap, Ownable, TokenCollector, EIP712 {
                 if (makingAmount > orderAvailableAmount) revert NotEnoughForFill();
                 takerTokenAmounts[i] = ((makingAmount * order.takerTokenAmount) / order.makerTokenAmount);
 
+                if (order.takerToken == address(weth)) {
+                    wethToPay += takerTokenAmounts[i];
+                }
```
pilagod, line 97: I think these block could be placed after zero amount check (line 99)

test/forkMainnet/LimitOrderSwap/GroupFill.t.sol
```diff
@@ -233,7 +215,8 @@ contract GroupFillTest is LimitOrderSwapTest {
         // scenario
         // maker tokens (input): 5 WETH, 8000 USDT
         // taker tokens (output): 1 ETH, 3 WETH, 8000 USDT
-        // profit for arbitrageur: 1 ETH/WETH
+        // should unwrap 4 ether and leave 1 WETH
```
pilagod, line 218: I think this case unwraps only 2 WETH (5 WETH - 3 WETH), and 1 ETH for order and another 1 ETH for arbitrageur's profit, right?

charlesjhongc: Yeah, you're right. Unwrap 3 and profit is 1.

charlesjhongc: Fixed in 0708ce1

contracts/LimitOrderSwap.sol
charlesjhongc: Fixed in 0708ce1

NIC619 (approved):

pilagod (approved):

PR #234, opened 2023-06-27
LO add ReentrancyGuard

charlesjhongc:


charlesjhongc:
@NIC619 This is raised by internal audit from our security team. No precise path is provided but it's [possible](https://github.com/consenlabs/tokenlon-contracts/blob/master/contracts/LimitOrderSwap.sol#L247-L251) for attacker for exploiting. Instead of managing code order, I think `ReentrancyGuard` is more suitable here.

PR #237, opened 2023-06-29
add shared contract base

charlesjhongc:


NIC619 (dismissed):

contracts/UniAgent.sol
```diff
@@ -5,7 +5,7 @@ import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 import { TokenCollector } from "./abstracts/TokenCollector.sol";
 import { EIP712 } from "./abstracts/EIP712.sol";
-import { Ownable } from "./abstracts/Ownable.sol";
+import { Ownable } from "./abstracts/Base.sol";
```
pilagod, line 8: How about importing `Ownable` from `./abstracts/Ownable.sol`?

contracts/abstracts/Base.sol
```diff
@@ -0,0 +1,34 @@
...
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import { Ownable } from "./Ownable.sol";
+import { Asset } from "../libraries/Asset.sol";
+import { Constant } from "../libraries/Constant.sol";
+
+/// @title Base Contract
+/// @author imToken Labs
+abstract contract Base is Ownable {
```
pilagod, line 13: Maybe it could have a more meaningful name, for example `TokenManager` or `TokenManagement`.

contracts/UniAgent.sol
charlesjhongc: It should be fixed. 369ef43.

contracts/abstracts/Base.sol
charlesjhongc: Renamed as `AdminManagement` in 80d671b

pilagod (approved):

NIC619 (approved):

PR #246, opened 2023-07-04
refactor foundry config

charlesjhongc:
Seems like `slither` doesn't work with non-default foundry output path. Also, remove default configs in file to make it clear.

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -1,19 +1,9 @@
 [profile.default]
 src = 'contracts'                                             # the source directory
-test = 'test'                                                 # the test directory
-out = 'artifacts'                                             # the output directory (for artifacts)
-cache = true                                                  # whether to cache builds or not
-force = false                                                 # whether to ignore the cache (clean build)
 solc_version = "0.8.17"
 optimizer = true                                              # enable or disable the solc optimizer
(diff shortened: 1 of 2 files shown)


foundry.toml
```diff
@@ -1,19 +1,9 @@
 [profile.default]
 src = 'contracts'                                             # the source directory
-test = 'test'                                                 # the test directory
-out = 'artifacts'                                             # the output directory (for artifacts)
-cache = true                                                  # whether to cache builds or not
-force = false                                                 # whether to ignore the cache (clean build)
```
NIC619, line 6: > Also, remove default configs in file to make it clear.

In my opinion removing default values would not make it clear as we can not memorize default values in our head. At least for the config values we care, we can leave them in the config file so we don't have to run `forge config` every time we want to know what the default value is.

charlesjhongc: Yeah, I agree some default value we care should be listed here. Perhaps only `cache` and `force`?

NIC619: Makes sense!

charlesjhongc: Adjusted 788db72

NIC619 (approved):

pilagod (approved):

PR #264, opened 2023-07-20
refactor if statement gas

charlesjhongc:


pilagod (approved):
It's interesting to see gas diff has no changes in this refactoring 🤣 From my experience the `&&` operator would take the right part into account only if the first expression resolves to `false`. But I'm not sure if this is also true for Solidity 🧐

NIC619:
This refactor should result in gas reduction?

NIC619:
> From my experience the && operator would take the right part into account only if the left expression resolves to `false`.

I think it is "... only if the left expression resolves to `true`"?

charlesjhongc:
Tested in local and yes the gas is reduced a little bit.
Screenshot 2023-07-24 at 3 49 54 PM

charlesjhongc:
@pilagod If you meant there's no gas diff comment, I think it's because the CI setting? Or do you mean there's no gas diff in your local test?

NIC619 (approved):

pilagod:
I meant gas diff in CI. Yap it seems have some errors generating the report when merging into branches other than `master` 🧐 Thanks for the report screenshot 🙌

pilagod (approved):

PR #266, opened 2023-07-20
refactor arithmetic check

charlesjhongc:


contracts/AllowanceTarget.sol
```diff
@@ -16,8 +16,11 @@ contract AllowanceTarget is IAllowanceTarget, Pausable, Ownable {
 
     constructor(address _owner, address[] memory trustedCaller) Ownable(_owner) {
         uint256 callerCount = trustedCaller.length;
-        for (uint256 i = 0; i < callerCount; ++i) {
+        for (uint256 i = 0; i < callerCount; ) {
             authorized[trustedCaller[i]] = true;
+            unchecked {
+                ++i;
+            }
```
pilagod, line 23: I feel that this is a bit too much for gas optimization. I would prefer trading this gas cost for coding convention and readability.

contracts/LimitOrderSwap.sol
```diff
@@ -82,7 +82,7 @@ contract LimitOrderSwap is ILimitOrderSwap, Ownable, TokenCollector, EIP712, Ree
         uint256[] memory takerTokenAmounts = new uint256[](orders.length);
         uint256 wethToPay;
         address payable _feeCollector = feeCollector;
-        for (uint256 i = 0; i < orders.length; ++i) {
+        for (uint256 i = 0; i < orders.length; ) {
```
pilagod, line 85: nit: Shall we cache the `orders.length` and also other loops in this PR?

contracts/AllowanceTarget.sol
NIC619: I feel like this pattern is more difficult to follow too. Unless documented properly, devs could misuse or break the pattern when copy-pasting it in the future.

charlesjhongc: Ok, let's skip this refactoring for now. It's kind of common for me but I do think it's more error-prone.

contracts/LimitOrderSwap.sol
charlesjhongc: Will do in another PR.

contracts/AllowanceTarget.sol
charlesjhongc: Fixed.

NIC619 (approved):

pilagod (approved):

charlesjhongc:
Fix array length cache in #268

PR #275, opened 2023-08-02
Revert "cache array length"

charlesjhongc:
Caching array length of a calldata array actually increase the gas consumption instead. Revert the previous change.


charlesjhongc:
Gas down after revert

Screenshot 2023-08-02 at 9 26 19 PM

NIC619 (approved):

NIC619:
I guess it depends on how long the array is?

pilagod (approved):

charlesjhongc:
@NIC619 It depends on the storage type.

PR #276, opened 2023-08-02
refactor TokenCollector

charlesjhongc:
Make the permit encoding consistent with v6 latest design.


test/utils/Permit2Helper.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.7.6;
+pragma abicoder v2;
```
pilagod, line 3: Long time no see the `abicoder` declaration since we have already migrated to v0.8 for a while 🤣

pilagod (approved):

test/utils/Permit2Helper.sol
charlesjhongc: Indeed! After being around v0.8 for a while, now I think v0.7 is really out of date.

contracts/utils/TokenCollector.sol
```diff
@@ -32,91 +33,51 @@ abstract contract TokenCollector {
...
-        uint256 amount,
-        bytes memory data
-    ) private {
-        if (data.length > 0) {
-            (bool success, bytes memory result) = token.call(abi.encodePacked(IERC20Permit.permit.selector, data));
+        if (src == Source.TokenlonSpender) {
+            ISpender(tokenlonSpender).spendFromUser(from, token, amount);
+            if (to != address(this)) {
+                IERC20(token).safeTransfer(to, amount);
+            }
```
NIC619, line 44: This is because we won't be deploying v5 `Spender` to a new version that supports `spendFromUserTo`? I see that there's `spendFromUserTo` in `ISpender` and `Spender`. Maybe we should rollback v5 `Spender` to the older version?

NIC619 (approved):

contracts/utils/TokenCollector.sol
charlesjhongc: Yeah, the spender won't be updated. I think rolling back is a good idea, will be in another PR.

PR #291, opened 2023-08-30
add tests for SignatureValidator

charlesjhongc:
`WalletBytes` is not standard EIP-1271 and never used in any places include MMSK so no tests were added for it.


test/SignatureValidator/EIP712.t.sol
```diff
@@ -0,0 +1,55 @@
...
+        uint256 r = 1;
+        bytes memory signature = abi.encodePacked(r, sigType);
+        // should have 33 bytes signature
+        assertEq(signature.length, 33);
+        vm.expectRevert("SignatureValidator#isValidSignature: length 65 or 97 required");
+        isValidSignature(vm.addr(userPrivateKey), digest, bytes(""), signature);
+    }
+
+    /// @dev old contracts still assert sigLength == 97 so has to support this format
+    /// @dev the extra bytes32 is not used at all
```
pilagod, line 33: nit: I think it would be clearer to have exactly one NatSpec tag for each type.

```suggestion
    /// @dev Old contracts still assert sigLength == 97 so has to support this format,
    ///            and the extra bytes32 is not used at all.
```

test/SignatureValidator/EIP712.t.sol
```diff
@@ -0,0 +1,55 @@
...
+    function testEIP712With97BytesSig() public {
+        (uint8 v, bytes32 r, bytes32 s) = vm.sign(userPrivateKey, digest);
+        bytes memory signature = abi.encodePacked(r, s, v, bytes32(0), sigType);
+        // standard ECDSA signature : 65 bytes
+        // extra bytes32 : 32 bytes
+        // signatureType : 1 byte
+        // total : 98 bytes
+        assertEq(signature.length, 98);
```
pilagod, line 41: Q: So the signature length should be 97 (mentioned in test name) or 98? 😂

test/mocks/MockZX1271Wallet.sol
```diff
@@ -0,0 +1,25 @@
...
+    bytes4 public constant ZX1271_MAGICVALUE = bytes4(keccak256("isValidWalletSignature(bytes32,address,bytes)"));
+
+    address public operator;
+
+    constructor(address _operator) {
+        operator = _operator;
+    }
+
+    function isValidSignature(bytes calldata _data, bytes calldata _signature) external view override returns (bytes4 magicValue) {
+        require(operator == ECDSA.recover(keccak256(_data), _signature), "MockERC1271Wallet: invalid signature");
```
pilagod, line 17: Q: Does the production version ZX1271Wallet share the same behavior of revert when signature is not valid?

test/SignatureValidator/Setup.t.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.7.6;
+pragma abicoder v2;
+
+import { Test } from "forge-std/Test.sol";
+import { SignatureValidator } from "contracts/utils/SignatureValidator.sol";
+import { IERC1271Wallet } from "contracts/interfaces/IERC1271Wallet.sol";
+import { ECDSA } from "@openzeppelin/contracts/cryptography/ECDSA.sol";
+
+contract TestSignatureValidator is SignatureValidator, Test {
```
NIC619, line 10: Maybe not inherit from `SignatureValidator` and deploy it as a standalone contract so the tests won't have unexpected behavior?

test/SignatureValidator/EIP712.t.sol
charlesjhongc: Hmm it depends on the definition of signature. The total length of `signature` passing into function `isValidSignature()` is 98 bytes (97 + 1 for signature type). I will change the name to 98 to avoid misunderstanding.

test/mocks/MockZX1271Wallet.sol
charlesjhongc: It varies from different implementations. One of two ZX1271 wallets in our production reverts when invalid (https://etherscan.io/address/0x3b938e9525e14361091ee464d8acec291b3cae50#code). Another one didn't upload their code to etherscan.

test/SignatureValidator/Setup.t.sol
charlesjhongc: Fixed 21e54ebc9e203b6b082fff497bbbf49fbb9a6e05

test/mocks/MockZX1271Wallet.sol
charlesjhongc: Added two regression to make sure it won't break anything 6fcccf2afdf7cac02c8aba2dbe174378bfe1c793

test/SignatureValidator/EIP712.t.sol
charlesjhongc: Fixed a978a7b2ece6e54f6cadc2fb20c3623625dd806f

pilagod (approved):

108356037 (approved):

charlesjhongc:
@NIC619 Do you have any other comments?

NIC619 (approved):
Forgot to re-approve

PR #295, opened 2023-09-11
refactor signatureValidator

charlesjhongc:


contracts/utils/SignatureValidator.sol
```diff
@@ -1,195 +1,91 @@
...
-            return isValid;
-        }
 
-        // Anything else is illegal (We do not return false because
-        // the signature may actually be valid, just not in a format
-        // that we currently support. In this case returning false
-        // may lead the caller to incorrectly believe that the
-        // signature was invalid.)
-        revert("SignatureValidator#isValidSignature: unsupported signature");
+        revert("SignatureValidator#isValidSignature: illegal signature");
```
NIC619, line 56: Maybe we can skip or replace `SignatureValidator#isValidSignature`?

contracts/utils/SignatureValidator.sol
```diff
@@ -1,195 +1,91 @@
...
+    } else if (signatureType == SignatureType.EIP712) {
+        // In ealier version, an extra bytes32 is packed with (v,r,s) but the field was never used.
+        // So the _sig was in 97 bytes before.
+        // In order to support standard 65 bytes ECDSA signature and legacy format, still spliting _sig here
+        // instead of passing _sig directly to OZ ECDSA lib.
```
NIC619, line 61: I would suggest:
```suggestion
        // To be backward compatible with previous signature format which has (v,r,s) arranged in the order (r,s,v) and with an extra 32 bytes padded in the end, here we just extract the (v,r,s) from the signature and ignore the rest.
```

contracts/utils/SignatureValidator.sol
```diff
@@ -1,195 +1,91 @@
...
-            }
-        }
-        return isValid;
-    }
+    // Anything else is illegal (We do not return false because
+    // the signature may actually be valid, just not in a format
+    // that we currently support. In this case returning false
+    // may lead the caller to incorrectly believe that the
+    // signature was invalid.)
+    revert("SignatureValidator: unsupported signature");
```
NIC619, line 90: This will never be reached because we check `signatureTypeRaw <= uint8(SignatureType.ZX1271)`?

contracts/utils/SignatureValidator.sol
```diff
@@ -1,195 +1,91 @@
...
+// 0xb0671381
+bytes4 constant ZX1271_MAGICVALUE = bytes4(keccak256("isValidWalletSignature(bytes32,address,bytes)"));
+
+enum SignatureType {
+    Illegal, // 0x00, default as illegal
+    Invalid, // 0x01
+    EIP712, // 0x02 standard EIP-712 signature
+    EthSign, // 0x03 signed using web3.eth_sign() or Ethers wallet.signMessage()
+    WalletBytes, // 0x04 DEPRECATED, never used
+    Standard1271, // 0x05 standard EIP-1271 wallet type
```
pilagod, line 20: nit: I would prefer calling it `EIP1271`

contracts/utils/SignatureValidator.sol
charlesjhongc: The `SignatureType .Invalid` will reach here.

charlesjhongc: Maybe can remove `SignatureType.Illegal` case?

contracts/utils/SignatureValidator.sol
charlesjhongc: Fixed in 2366b87

contracts/utils/SignatureValidator.sol
NIC619: By looking at the code it seems that OZ lib expects the order `v,r,s`: `ECDSA.recover(_hash, v, r, s);`

while we expect the order in the signature to be `r,s,v`?
```solidity
bytes32 r = LibBytes.readBytes32(_sig, 0);
bytes32 s = LibBytes.readBytes32(_sig, 32);
uint8 v = uint8(_sig[64]);
```

contracts/utils/SignatureValidator.sol
NIC619: Since `Illegal` also serves the purpose of representing the default value of enum type, maybe we can remove `Invalid` instead? (This final revert is implicitly the case for `Invalid` anyway.)

contracts/utils/SignatureValidator.sol
charlesjhongc: Oh I mean this function `function recover(bytes32 hash, bytes memory signature)`. The layout of `bytes signature` is (r,s,v).

contracts/utils/SignatureValidator.sol
charlesjhongc: The goal is not changing the enum order. But if you're talking about `if (signatureType == SignatureType.Illegal)` case, yeah I was thinking about removing this.

contracts/utils/SignatureValidator.sol
NIC619: I see what you mean. So we are just ditching the rest of the bytes beyond 65th byte.
```suggestion
        // To be backward compatible with previous signature format which has an extra 32 bytes padded in the end, here we just extract the (r,s,v) from the signature and ignore the rest.
```

charlesjhongc: Updated in 8bcb604

contracts/utils/SignatureValidator.sol
NIC619: > yeah I was thinking about removing this.

Yeah that would make sense. Though I'm still not fully understand the difference between `Illegal` and `Invalid`.🧐

contracts/utils/SignatureValidator.sol
NIC619: @charlesjhongc Should the prefix be updated?

contracts/utils/SignatureValidator.sol
charlesjhongc: After discussion, fixed in 472d85b0f54876e716461f8633e9cbcc3da1e16e

contracts/utils/SignatureValidator.sol
charlesjhongc: Updated in 472d85b0f54876e716461f8633e9cbcc3da1e16e

NIC619 (approved):

pilagod (approved):

PR #303, opened 2023-12-05
gas optimization: leave one wei feature

alex0207s:
For gas optimization purposes, we retain 1 wei of each token in our contract(GS and SOS). This will impact the initial swap of every token, as 1 wei will be left in our contract.

- Implement the logic for leaving 1 wei in `GenericSwap.sol`
- Implement the logic for leaving 1 wei in `SmartOrderStrategy.sol`

### DIFF
--- a/test/forkMainnet/GenericSwap.t.sol
+++ b/test/forkMainnet/GenericSwap.t.sol
@@ -157,17 +157,17 @@ contract GenericSwapTest is Test, Tokens, BalanceUtil, Permit2Helper, SigHelper 
         Snapshot memory makerMakerToken = BalanceSnapshot.take({ owner: address(mockStrategy), token: gsData.makerToken });
 
         uint256 actualOutput = 900;
-        uint256 realChangedInGas = actualOutput - 1;
+        uint256 realChangedInGS = actualOutput - 1; // leaving 1 wei in GS
 
         // 800 < 900 < 1000
         mockStrategy.setOutputAmountAndRecipient(actualOutput, payable(address(genericSwap)));
         vm.expectEmit(true, true, true, true);
-        emit Swap(getGSDataHash(gsData), gsData.maker, taker, taker, gsData.takerToken, gsData.takerTokenAmount, gsData.makerToken, realChangedInGas);
+        emit Swap(getGSDataHash(gsData), gsData.maker, taker, taker, gsData.takerToken, gsData.takerTokenAmount, gsData.makerToken, realChangedInGS);
         vm.prank(taker);
         genericSwap.executeSwap(gsData, defaultTakerPermit);
 
         takerTakerToken.assertChange(-int256(gsData.takerTokenAmount));
-        takerMakerToken.assertChange(int256(realChangedInGas));
+        takerMakerToken.assertChange(int256(realChangedInGS));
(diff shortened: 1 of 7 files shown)


test/forkMainnet/SmartOrderStrategy/Setup.t.sol
```diff
@@ -40,6 +40,11 @@ contract SmartOrderStrategyTest is Test, Tokens, BalanceUtil {
             setERC20Balance(tokenList[i], genericSwap, 10000);
         }
 
+        deal(USDC_ADDRESS, address(smartOrderStrategy), 1 wei);
+        deal(USDT_ADDRESS, address(smartOrderStrategy), 1 wei);
+        deal(WETH_ADDRESS, address(smartOrderStrategy), 1 wei);
+        deal(WBTC_ADDRESS, address(smartOrderStrategy), 1 wei);
```
108356037, line 46: Suggestion: Try testing if leaving one wei works on first swap instead  of dealing to contract directly.

charlesjhongc: Same with David.

alex0207s: I prefer sticking with the current approach for these tests, and instead, introducing new test scenarios to verify that leaving 1 wei works on the first swap. Any thoughts on that?

test/forkMainnet/GenericSwap.t.sol
```diff
@@ -156,16 +157,17 @@ contract GenericSwapTest is Test, Tokens, BalanceUtil, Permit2Helper, SigHelper
         Snapshot memory makerMakerToken = BalanceSnapshot.take({ owner: address(mockStrategy), token: gsData.makerToken });
 
         uint256 actualOutput = 900;
+        uint256 realChangedInGas = actualOutput - 1;
```
charlesjhongc, line 160: Gas or GS?

test/forkMainnet/GenericSwap.t.sol
```diff
@@ -183,12 +185,21 @@ contract GenericSwapTest is Test, Tokens, BalanceUtil, Permit2Helper, SigHelper
...
-        emit Swap(getGSDataHash(gsData), gsData.maker, taker, taker, gsData.takerToken, gsData.takerTokenAmount, gsData.makerToken, gsData.makerTokenAmount);
+        emit Swap(
+            getGSDataHash(gsData),
+            gsData.maker,
+            taker,
+            taker,
+            gsData.takerToken,
+            gsData.takerTokenAmount,
+            gsData.makerToken,
+            gsData.makerTokenAmount - 1
```
charlesjhongc, line 196: I think we can have `expectedOut = gsData.makerTokenAmount - 1` for these assertion.

charlesjhongc (commented):
Overall I think both ways are fine as long as cases with and without 1wei in contracts are all covered in tests. However, cases with amount replacement are more important to include the 1wei context since the way it affects the ratio calculation.

test/forkMainnet/GenericSwap.t.sol
alex0207s: Good idea, but maybe call it `realChangedInGS` for consistency.

charlesjhongc (approved):
LGTM

PR #310, opened 2024-04-30
Fix v6.0.1

alex0207s:
- fix typo in RFQ contract
- add a parameter (`salt`) to `Swap` event in GenericSwap contract
- change handling logic for makerToken as ETH in RFQ contract
- change the input token ratio calculation method in SmartOrderStrategy contract
- remvoe `uniAgent.sol` contract

### Why do we need to adjust the handling logic for makerToken as ETH in the RFQ contract?

Prior to this PR, when makerToken is ETH, RFQ directly unwraps WETH into ETH. This isn't ideal, especially when RFQ is only a segment of the entire swap path. If RFQ unwraps WETH into ETH directly, it requires the subsequent swapping process to wrap ETH back into WETH, incurring additional steps and costs.

With this PR, we don't unwrap WETH directly. WETH is only unwrapped to ETH when the makerToken is ETH. This could avoid additional steps or costs associated with handling WETH.

### Why do we need to adjust the calculation method for a input token ratio?
Prior to this PR, the ratio value was expressed in ten-thousandths (0.01%), which was not precise enough for some specific protocols(e.g. [dsspsm](https://etherscan.io/address/0x89B78CfA322F6C5dE0aBcEecab66Aee45393cC5A#code)). 

With this PR, the ratio is now determined by both the numerator and the denominator, allowing for greater flexibility and accuracy in expressing the ratio.

### Why do we remove `uniAgent.sol`?
Due to a backend integration issue, the current version of uniAgent.sol is not sufficient to replace GenericSwap.sol and SmartOrderStrategy.sol when the swap only occurs in the Uniswap pool.


NIC619:
> fix handling logic for makerToken as ETH in RFQ contract

This does not feel like a "fix". This is changing how we treat WETH as maker token:
- before this PR: ETH == WETH
    - No ETH as maker token allowed; WETH as maker token means ETH as maker token, we converts it to ETH and send it to recipient
- after this PR: ETH != WETH 
    - WETH as maker token is the same as other ERC20 tokens, we will not convert it to ETH anymore; ETH as maker token means we will convert WETH to ETH and send it to recipient

NIC619:
Please add a readme or a natspec/comment for how we handle different maker tokens, to help others better understand how it works.

```
In RFQ,
- if maker token is ETH, ...
- if maker token is WETH, ...
- if maker token is normal ERC20, ...
```

foundry.toml
```diff
@@ -11,6 +11,7 @@ fs_permissions = [
     { access = "read", path = "./test/utils/config/" },
     { access = "read", path = "./test/utils/payload/" },
 ]
+via-ir = true
```
NIC619, line 14: Please do not just embed this kind of change in an unrelated commit. Open a PR to list reasons why you want/support this change and start a discussion there.

Please answer the following:
- How much does it impact compile time?
- Would contract verification still work in `tokenlon-contract-private` repo? [ref](https://twitter.com/PaulRBerg/status/1737169159052468479)
- Would it impact the scripts we run in `tokenlon-contract-private` repo? [ref](https://hardhat.org/hardhat-runner/docs/reference/solidity-support#support-for-ir-based-codegen): `> Since Hardhat works much better when the optimizer is disabled, we don't completely support the viaIR option yet.`
- Are there any other caveats of enabling IR? These caveats need to be documented in readme before we enable this change.

NIC619: This does not mean I'm against enabling `via-ir` but rather we need more study and feedbacks before we enabling this kind of change. Otherwise we might waste more time debugging in the future.

contracts/RFQ.sol
```diff
@@ -133,13 +133,19 @@ contract RFQ is IRFQ, Ownable, TokenCollector, EIP712 {
...
-        // collect makerToken from maker to this
+        // collect makerToken from maker to this contract
         uint256 makerSettleAmount = _rfqOffer.makerTokenAmount;
         if (_rfqTx.takerRequestAmount != _rfqOffer.takerTokenAmount) {
             makerSettleAmount = (_rfqTx.takerRequestAmount * _rfqOffer.makerTokenAmount) / _rfqOffer.takerTokenAmount;
         }
         if (makerSettleAmount == 0) revert InvalidMakerAmount();
-        _collect(_rfqOffer.makerToken, _rfqOffer.maker, address(this), makerSettleAmount, _makerTokenPermit);
+        // if the makerToken is ETH, we collect WETH from the maker to this contract
+        // if the makerToken is a ERC20 token, we collect that ERC20 token from maker to this contract
```
NIC619, line 143:
```suggestion
        // if the makerToken is a ERC20 token (including WETH), we collect that ERC20 token from maker to this contract
```

contracts/RFQ.sol
```diff
@@ -150,9 +156,9 @@ contract RFQ is IRFQ, Ownable, TokenCollector, EIP712 {
         }
 
         {
-            // determine if WETH unwrap is needed, send out ETH if makerToken is WETH
+            // determine if WETH unwrap is needed, send out ETH if makerToken is ETH
```
NIC619, line 159:
```suggestion
            //  unwrap WETH and send out ETH if makerToken is ETH
```

contracts/RFQ.sol
```diff
@@ -150,9 +156,9 @@ contract RFQ is IRFQ, Ownable, TokenCollector, EIP712 {
...
             address makerToken = _rfqOffer.makerToken;
-            if (makerToken == address(weth)) {
+            if (makerToken.isETH()) {
                 weth.withdraw(makerSettleAmount);
                 makerToken = Constant.ETH_ADDRESS;
```
NIC619, line 163:
```suggestion
```

NIC619 (approved):

contracts/interfaces/ISmartOrderStrategy.sol
```diff
@@ -16,8 +16,9 @@ interface ISmartOrderStrategy is IStrategy {
     struct Operation {
         address dest;
         address inputToken;
-        uint128 inputRatio;
-        uint128 dataOffset;
+        uint256 numerator;
+        uint256 denominator;
+        uint256 dataOffset;
```
alex0207s, line 21: To save gas, replace a uint128 with a uint256. Using uint128 would incur read/write overhead and consume more gas.

NIC619: Can `numerator` and `denominator` be packed into uint256?

contracts/interfaces/ISmartOrderStrategy.sol
```diff
@@ -16,8 +16,9 @@ interface ISmartOrderStrategy is IStrategy {
     struct Operation {
         address dest;
         address inputToken;
-        uint128 inputRatio;
-        uint128 dataOffset;
+        uint256 numerator;
+        uint256 denominator;
```
NIC619, line 20: Suggest rename: `ratioNumerator` and `ratioDenominator`

contracts/SmartOrderStrategy.sol
```diff
@@ -79,15 +84,20 @@ contract SmartOrderStrategy is ISmartOrderStrategy, AdminManagement {
...
             assembly {
                 mstore(add(_data, _dataOffset), inputTokenBalance)
             }
         }
 
-        (bool success, bytes memory result) = _dest.call{ value: _value }(_data);
+        bool success;
+        bytes memory result;
+        if (_value == 0) (success, result) = _dest.call(_data);
+        else (success, result) = _dest.call{ value: _value }(_data);
```
NIC619, line 99: What are the pros and cons of this change?

test/forkMainnet/SmartOrderStrategy/AMMs.t.sol
```diff
@@ -117,7 +119,8 @@ contract AMMsTest is SmartOrderStrategyTest {
         operations[0] = ISmartOrderStrategy.Operation({
             dest: UNISWAP_SWAP_ROUTER_02_ADDRESS,
             inputToken: defaultInputToken,
-            inputRatio: Constant.BPS_MAX, // BPS_MAX indicate the input amount will be replaced by the actual balance
+            numerator: Constant.BPS_MAX, // BPS_MAX indicate the input amount will be replaced by the actual balance
+            denominator: 10000,
```
NIC619, line 123:
```suggestion
            numerator: 10000, // same numerator and denominator indicate the input amount will be replaced by the actual balance
            denominator: 10000,
```

test/forkMainnet/SmartOrderStrategy/AMMs.t.sol
```diff
@@ -37,7 +37,8 @@ contract AMMsTest is SmartOrderStrategyTest {
         operations[0] = ISmartOrderStrategy.Operation({
             dest: UNISWAP_SWAP_ROUTER_02_ADDRESS,
             inputToken: defaultInputToken,
-            inputRatio: 0, // zero ratio indicate no replacement
+            numerator: 0, // zero ratio indicate no replacement
+            denominator: 10000,
```
NIC619, line 41: All `denominator` in zero ratio cases can be set to `0`:
```suggestion
            numerator: 0, // zero ratio indicate no replacement
            denominator: 0,
```

contracts/SmartOrderStrategy.sol
```diff
@@ -79,15 +84,20 @@ contract SmartOrderStrategy is ISmartOrderStrategy, AdminManagement {
             }
 
             // calculate input amount if ratio should be applied
-            if (_inputRatio != Constant.BPS_MAX) {
-                inputTokenBalance = (inputTokenBalance * _inputRatio) / Constant.BPS_MAX;
+            // we provide a _numerator and a _denominator to decide a ratio
+            if (_numerator != _denominator) {
+                inputTokenBalance = (inputTokenBalance * _numerator) / _denominator;
```
NIC619, line 89: Add a test for `_denominator` being `0` while `_numerator > 0`

contracts/interfaces/ISmartOrderStrategy.sol
alex0207s: They can be packed into a uint256, but this will consume more gas.

### Gas cost of the `executeStrategy` function when we declare `numerator` and `denominator` as uint256 separately:
- min: 23253
- avg: 126523
- median: 84289
- max: 610589

### Gas cost of the `executeStrategy` function when we pack `numerator` and `denominator` into a uint256:

- min: 23253
- avg: 126662
- median: 84450
- max: 610929

test/forkMainnet/SmartOrderStrategy/AMMs.t.sol
alex0207s: ACK

test/forkMainnet/SmartOrderStrategy/AMMs.t.sol
alex0207s: ACK

contracts/SmartOrderStrategy.sol
alex0207s: I initially made this change to save gas, but after re-running it, I found that it actually consumes more gas. Maybe I used the wrong baseline for comparison before. I will remove this change

contracts/SmartOrderStrategy.sol
alex0207s: ACK

NIC619 (approved):

PR #336, opened 2024-12-16
feature: decouple strategyData from  GenericSwapData

alex0207s:
`strategyData` is time-sensitive, and including it in the `GenericSwapData` struct would make the entire order time-restricted. By decoupling `strategyData`, the order becomes more flexible, ensuring that the user's intent to exchange tokens remains unchanged while avoiding the need to repeatedly sign the same exchange intent with different `strategyData`.


test/forkMainnet/GenericSwap.t.sol
```diff
@@ -50,6 +49,7 @@ contract GenericSwapTest is Test, Tokens, BalanceUtil, Permit2Helper, SigHelper
     uint24[] defaultV3Fees = [3000];
     bytes defaultTakerPermit;
     bytes alicePermit;
+    bytes swapData;
```
NIC619, line 52: nit: maybe rename it to `strategyData` to avoid confusion as the first argument of `executeSwap*` is also `swapData`.

NIC619 (approved):

